PR #2, opened 2016-04-22
OVN docker image built from source

huikang:
- simplify ovn setup script for container (due to the fs namespace isolation
  of docker container)
- OVN image is built by compiing OVS from git


huikang:
Hi, Lei,
some improvements on OVN docker images and the setup script. These steps are necessary for integrating rally workload with docker-based OVN scale test. Thanks. - Hui

PR #5, opened 2016-04-24
Enable system-wide installation

l8huang:
Load plugins from code directly without command line parameter or environment variable.


huikang:
LGTM
Tested and it works very well. - Hui

PR #11, opened 2016-05-06
Add port binding rally workload file

huikang:
- register docker sandbox in rally
- Working port bind
- Add create and list port workload file

### DIFF
--- a/ansible/roles/rally/templates/create_and_bind_ports.json.j2
+++ b/ansible/roles/rally/templates/create_and_bind_ports.json.j2
@@ -1,0 +1,36 @@
+{
+    "version": 2,
+    "title": "Create and bind port",
+    "subtasks": [{
+        "title": "Create and bind port",
+        "workloads": [{
+            "name": "OvnNetwork.create_and_bind_ports",
+            "args": {
+                "network_create_args": {
+                    "amount": {{ network_number }},
+                    "batch": 1,
+                    "start_cidr": "{{ network_start_cidr }}",
+                    "install_method": "docker",
+                    "physical_network": "providernet"
+                },
+                "port_create_args" : {
+                    "install_method": "docker",
+                    "batch": 1
+                },
+                "ports_per_network": {{ ports_per_network }},
+                "port_bind_args": {
+                    "wait_up": true,
+                    "install_method": "docker"
+                }
+            },
+            "runner": {
+                "type": "serial","times": 1},
+            "context": {
+               "ovn_multihost" : {
+                    "controller": "ovn-controller-node"
+                },
+                "sandbox":{ "tag": "ToR1"}
+            }
+        }]
+    }]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ doc/source/_build/
 *.swp
 
 *~
+site.retry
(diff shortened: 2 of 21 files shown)


huikang:
Port binding workload is done. Now we can stress the southbound database and the ovn-controller. - Hui

l8huang:
Seems like 'install_method' is a problem now, it is coming up everywhere :)

'docker' and 'bare-metal' are different deployment methods in my mind, why not move it into deployment config, e.g. add a 'method'  to [ovn-multihost-deployment.json.j2](https://github.com/openvswitch/ovn-scale-test/blob/master/ansible/roles/rally/templates/ovn-multihost-deployment.json.j2#L7), record it in Deployment object in DB, then initialize ovsclients with the deployment 'method' at [rally_ovs.plugins.ovs.scenario.OvsScenario](https://github.com/openvswitch/ovn-scale-test/blob/master/rally_ovs/plugins/ovs/scenario.py#L38). In this way, the 'install_method' is only set once while deploying, all of tasks can follow it without additional configuration. 

In Rally's way, 'install_method' of 'docker' is better to be done with a deployment engine plugin, like 'OvnSandboxControllerEngine' and 'OvnSandboxFarmEngine'. You can wrap ansible commands mentioned in [README:Deploying OVN Emulation](https://github.com/huikang/ovn-scale-test/blob/ansible-docker-rally-port-binding/ansible/README.rst#deploying-ovn-emulation)  and [README:Registerring with Rally](https://github.com/huikang/ovn-scale-test/blob/ansible-docker-rally-port-binding/ansible/README.rst#registerring-with-rally) into the engine plugin.

ansible/roles/rally/templates/create_and_bind_ports.json.j2
```diff
@@ -0,0 +1,36 @@
+{
```
l8huang: It's better to keep these task config files under [samples](https://github.com/huikang/ovn-scale-test/tree/ansible-docker-rally-port-binding/samples) folder.

huikang:
Thanks for the suggestions. I think it would be better to split this patch into two smaller ones: one for the install_method and the other for the port_binding. I will create another PR and leave this as a reference.

huikang:
Lei, I was trying to add install_method as a [field to the deployment database](https://github.com/huikang/ovn-scale-test/blob/change-install-method/rally_ovs/plugins/ovs/deployment/engines/ovn_sandbox_controller.py#L133) as you suggested.

Later when creating the OvsClient, we can use `self._controller_clients = ovsclients.Clients(cred, install_method)`

My question is how can I retrieve this install_method when creating the [ovsclient](https://github.com/openvswitch/ovn-scale-test/blob/master/rally_ovs/plugins/ovs/scenario.py#L38).

Something like cred = v["install_method"] gives me compile error.

Thanks.

l8huang:
Deployment object has a field [config](https://github.com/openstack/rally/blob/master/rally/common/db/sqlalchemy/models.py#L80), it contains the config file's content used to create the deployment, so you can retrieve install method from it.

huikang:
To add `install_method` field into the deployment object, do I need to call `add_resource` like you did for [resourcetype.CONTROLLER](https://github.com/openvswitch/ovn-scale-test/blob/master/rally_ovs/plugins/ovs/deployment/engines/ovn_sandbox_controller.py#L127)?

I think I figure it out: dep.config is an accessible list. Thanks for your suggestion and will update a new PR.

huikang:
Merged as smaller PRs

PR #15, opened 2016-05-19
Add port binding rally workload file

huikang:
- register docker sandbox in rally
- create,  list, and bind port workload template file


huikang:
The last part split from #11

l8huang:
Cool!

PR #20, opened 2016-06-10
Fix ansible documentation

mestery:
The name of the json file used to create the emulated sandboxes has
changed.

### DIFF
--- a/ansible/README.rst
+++ b/ansible/README.rst
@@ -173,7 +173,9 @@ Register emulated sandboxes in the rally database
 
 ::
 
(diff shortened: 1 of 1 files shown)


huikang:
Hi, @mestery Thanks for fixing this. Actually the format of the create_sandbox file should be `create-sandbox-{HOSTNAME}.json`. You emulation host has the hostname of ovn-rally, therefore you got that json file.

The suffix hostname is appended by [this ansible template task](https://github.com/openvswitch/ovn-scale-test/blob/master/ansible/roles/rally/tasks/config.yml#L54).

mestery:
Done, I fixed it @huikang.

PR #29, opened 2016-06-15
Cleanup the CI/CD scripts

mestery:
This cleans things up such that we no longer clone the ovn-scale-test
repository (this code now lives in this repository). We also no longer
create a top-level directory but simply run from the ci directory.


huikang:
LGTM.
Cloning ovn-scale-test from github is unnecessary.

PR #30, opened 2016-06-16
Stop hardcoding the ubuntu user

mestery:
Determine what user to run as dynamically.


huikang:
Does that mean any user can run this CI script and launch docker container, as long as the user is in the sudo list?

mestery:
@huikang Yes, what was in there assumed the user was "ubuntu", which was not working on our internal CI.

huikang:
gotcha.

huikang:
If the change is tested in internal CI and success, I would like to merge it.

mestery:
@huikang It is tested and should be mergable!

PR #31, opened 2016-06-16
Cleanups to the CI/CD automation

mestery:
- Add an $OVNSUDO to allow docker commands to run with sudo
- Run "apt-get update" in prepare.sh early
- Ensure the root user's ssh key is added to authorized_keys for the root user


mestery:
@huikang Need this one as it cleans up some bad assumptions that were made. With this, I can run the CI on a fresh host and things work.

huikang:
Merged. We need installing the independency for a fresh host.

PR #32, opened 2016-06-16
Use $OVNSUDO when building docker containers in ansible

mestery:
This is needed when running on hosts where the user running the scripts does
not have access to the docker daemon.


mestery:
@huikang One more fix.

PR #37, opened 2016-06-17
Fix one instance of missing $OVNSUDO

mestery:


mestery:
Ninja merging simple fix.

PR #43, opened 2016-06-20
Allow root ssh logins from the local IP and docker subnet

mestery:
Also cleanup some previous debug from prior commits.


mestery:
Invalid.

Issue #61, opened 2016-06-30
ovn-rally is still running when running cleanup


mestery:
Running this:

```
sudo /usr/local/bin/ansible-playbook -i $OVN_DOCKER_HOSTS ansible/site.yml -e @$OVN_DOCKER_VARS -e action=clean
```

leaves ovn-rally running:

```
ubuntu@ovn-rally-2:~/ovn-scale-test/ci$ docker ps
CONTAINER ID        IMAGE                  COMMAND               CREATED             STATUS              PORTS               NAMES
cf5db6185290        ovn-scale-test-rally   "/usr/sbin/sshd -D"   8 minutes ago       Up 8 minutes                            ovn-rally
ubuntu@ovn-rally-2:~/ovn-scale-test/ci$ 
```

mestery:
Thanks @huikang!

PR #62, opened 2016-06-30
Fix ovn-rally cleanup task

huikang:
If deployer user is root, we do not need to remove the user


huikang:
Fix issue #61

PR #66, opened 2016-07-13
Collect raw results from scale runs.

jayhawk87:
Raw results are needed for determining performance
regressions via hypothesis testing.


jayhawk87:
meh... still not correct - please ignore

Issue #69, opened 2016-07-15
Add error checking to the CI tools


mestery:
Currently, the ovn-scale-test CI scripts in the ci/ directory do not handle errors if things fail, which would allow the underlying CI system to conclude the run was in error.

huikang:
@mestery we need some code to check the results of rally-ovs. The travis-ci has reported wrong results several times, e.g., https://travis-ci.org/openvswitch/ovn-scale-test/builds/146386945#L5947

PR #70, opened 2016-07-15
Add error handling

mestery:
Add some initial error handling into the CI scripts

Closes Issue #69

### DIFF
--- a/ci/docker-rally.sh
+++ b/ci/docker-rally.sh
@@ -4,6 +4,9 @@
 XTRACE=$(set +o | grep xtrace)
 set -o xtrace
 
+# Track failures
+FAILED=0
+
 OVS_REPO=${1:-https://github.com/openvswitch/ovs.git}
 OVS_BRANCH=${2:-master}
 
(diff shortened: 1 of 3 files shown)


mestery:
This needs a bit more work, but it's a start. Seeing how the travis-ci job run looks now.

ci/docker-rally.sh
```diff
@@ -4,6 +4,9 @@
 XTRACE=$(set +o | grep xtrace)
 set -o xtrace
 
+# Track failures
+FAILED=0
```
huikang, line 8: Is this variable used to indicate how many tests the CI job fails?

mestery, line 8: It's mainly used to detect if a specific command in docker-rally.sh fails. If run fails, we want to ensure we run cleanup, but we need to remember run failed so we can return that to travis-ci. Make sense?

ci/docker-rally.sh
```diff
@@ -12,13 +15,15 @@ echo "OVS_REPO=${OVS_REPO} OVS_BRANCH=${OVS_BRANCH}"
 # A combined script to run all the things
 
 # Prepare the environment
-./prepare.sh
+./prepare.sh || FAILED=$(( $FAILED + 1 ))
 
 # Run the testsuite
-./scale-run.sh $OVS_REPO $OVS_BRANCH
+./scale-run.sh $OVS_REPO $OVS_BRANCH || FAILED=$(( $FAILED + 1 ))
```
huikang, line 21: With the PR, shall we avoid calling exit in the other script, e.g., https://github.com/openvswitch/ovn-scale-test/blob/master/ci/scale-run.sh#L23

mestery, line 21: Yes, that should be a return instead of an exit. I will fix it.

huikang, line 21: thanks

huikang:
@mestery is the PR ready for review or more to add? Thanks.

mestery:
@huikang I think so, I plan to add more error backout in the other scripts, but this can merge as-is now. Thanks!

PR #72, opened 2016-07-18
Add configurable configure flags.

jayhawk87:
In cases where we want to test different configurations, having
configurable configure flags makes sense.


jayhawk87:
Ugh... this is missing some changes, will resubmit after I update

PR #73, opened 2016-07-18
Add configurable configure flags.

jayhawk87:
In cases where we want to test different configurations, having
configurable configure flags makes sense.

### DIFF
--- a/ci/scale-run.sh
+++ b/ci/scale-run.sh
@@ -9,6 +9,7 @@ source ovn-scale.conf
 
 OVS_REPO=$1
 OVS_BRANCH=$2
+CONFIG_FLAG=$3
 
 function check_container_failure {
     sleep 5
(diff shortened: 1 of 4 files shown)


ci/scale-run.sh
```diff
@@ -9,6 +9,7 @@ source ovn-scale.conf
 
 OVS_REPO=$1
 OVS_BRANCH=$2
+CONFIG_FLAG=$3
```
mestery, line 12: You missed one more step: you need to modify "docker-rally.sh" since that is what the CI calls.

mestery:
@huikang Any idea why the travis-ci never ran for this PR?

mestery:
@jayhawk87 I noticed that PR72 had a travis-ci run, but this one didn't. Once you update to address my comments I'll be curious to see if travis-ci picks this up and runs.

huikang:
@mestery I think the travis-CI executed and passed.

PR #75, opened 2016-07-18
Correct return's from non-functions to be exits

mestery:
Commit 3a2272f1148dd8cac3d387c061c3143dd639cb17 added some error handling,
and also changed some "exit" values to "return" values in scale-run.sh.
However, since those were being done outside a funciton, this resulted
in incorrect bash syntax as below:

Deploying failed, exiting
- return 1
  ./scale-run.sh: line 44: return: can only `return' from a function or sourced script
- popd

This commits changes the "return" values back to "exit" values.

### DIFF
--- a/ci/scale-run.sh
+++ b/ci/scale-run.sh
@@ -21,7 +21,7 @@ function check_container_failure {
         for failed in ${failed_containers}; do
             docker logs --tail all ${failed}
         done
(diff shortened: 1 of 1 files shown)


mestery:
@huikang Your eyes here as well when you have time.

PR #81, opened 2016-07-19
Fix a few nits in docker-rally.sh

jayhawk87:
Need to re-set xtrace after prepare.sh and rather than
return with the status, we need to exit with it.


mestery:
LGTM, waiting on travis-ci before merging.

PR #82, opened 2016-07-19
Fix a typo in scale-run.sh

jayhawk87:
Configure flags use configflags, not configflag


huikang:
Shall we fix these [1][2] as well?

[1] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/scale-run.sh#L12
[2] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/docker-rally.sh#L12

jayhawk87:
While not strictly necessary, it's probably a good idea - let me spin another patch

PR #84, opened 2016-07-20
Actually test real scale in create_and_bind_ports

jayhawk87:
Add ability to change port create batch size and
set it to 50 for ci.  Update ports_per_network to 500.

### DIFF
--- a/ansible/roles/rally/templates/create_and_bind_ports.json.j2
+++ b/ansible/roles/rally/templates/create_and_bind_ports.json.j2
@@ -13,7 +13,7 @@
                     "physical_network": "providernet"
                 },
                 "port_create_args" : {
-                    "batch": 1
+                    "batch": {{ port_created_batch_size }}
                 },
                 "ports_per_network": {{ ports_per_network }},
                 "port_bind_args": {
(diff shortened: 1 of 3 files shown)


ansible/roles/rally/templates/create_and_bind_ports.json.j2
```diff
@@ -13,7 +13,7 @@
                     "physical_network": "providernet"
                 },
                 "port_create_args" : {
-                    "batch": 1
+                    "batch": {{ port_create_batch_size }}
```
mestery: The CI failed because this should be "ports_created_batch_size", not "port_create_batch_size".

ci/ansible/all.yml
```diff
@@ -36,7 +36,8 @@ ovn_number_chassis: 5
 ########################
 network_start_cidr: "172.16.201.0/24"
 network_number: "5"
-ports_per_network: "50"
+ports_per_network: "500"
```
mestery, line 39: Do we really want to default this to 500? We should allow for this to be configured on the commandline, or have our downstream CI just modify all.yml before running.

jayhawk87, line 39: You are right that we should make this configurable, but I wasn't quite sure how to do that quickly last night ...

mestery:
I'll take this PR over as @jayhawk87 is swamped in a few other things today.

jayhawk87:
thanks @mestery

Issue #54, opened 2016-06-23
rally Dockerfile does not take effect from committed change


huikang:
Problem description: if a commit change anything in rally-ovs, the rally Dockerfile can not receive the new update. This is because rally Dockerfile clones the repo form upstream.

Then a following PR is needed to fix this issue, e.g., #53 .

One possible solution is to generate Dockerfile dynamically, using jinja2 tempate.

mestery:
@huikang Is this still relevant now that #53 has merged?

huikang:
@mestery Yes, this is because rally Dockerfile contains the link to this ovn-scale-test repo. We should fix this later.
But, lets keep this as an open issue so it will not get lost. Thanks.

huikang:
Lets keep this issue open for some time to ensure it works for everyone

PR #87, opened 2016-07-21
copy ovn-scale-test to ovn-rally container from local directory

huikang:
Without this patch, change to rally_ovs will not effect in the
travis-ci

### DIFF
--- a/ansible/docker/rally/Makefile
+++ b/ansible/docker/rally/Makefile
@@ -1,4 +1,7 @@
 image_name=ovn-scale-test-rally
 
 image:
+	tar cvzf /tmp/ovn-scale-test.tgz ../../../../ovn-scale-test
+	tar xvzf /tmp/ovn-scale-test.tgz
 	${OVNSUDO} docker build --force-rm=true -t ${image_name} .
+	rm -rf ./ovn-scale-test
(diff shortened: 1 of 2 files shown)


huikang:
address #54

ansible/docker/rally/Makefile
```diff
@@ -1,4 +1,7 @@
 image_name=ovn-scale-test-rally
 
 image:
+	tar cvzf /tmp/ovn-scale-test.tgz ../../../../ovn-scale-test
```
mestery, line 4: I'm not quite following what these changes are doing, can you explain? What I think is happening is this:
1. You are tar'ing up the existing ovn-scale-test directory, then untar'ing it in /tmp
2. the "docker build" command is using this directory from /tmp to build
3. You then rm the directory from /tmp

Is that right?

huikang:
@mestery your understanding is correct.
This PR is to facilitate any patch containing changes made to rally_ovs. Chances are there will be commits that modify `ovn-scale-test/rally_ovs` directory.
If using `git clone` in the Dockerfile; it always clones rally_ovs from upstream master rather than local branch.

mestery:
OK, thanks @huikang! This looks good then.

PR #88, opened 2016-07-21
Ensure prepare.sh works for hosts with a bond interface

mestery:
Found while testing in SoftLayer, where physical hosts have bonds
instead of individual network interfaces.


mestery:
@huikang Found this in testing today.

PR #89, opened 2016-07-21
Correctly setup docker-ovn-hosts on hosts with bond interfaces

mestery:


mestery:
@huikang One more bond fix.

PR #91, opened 2016-07-22
Fix rally_ovs workload in ci job

huikang:
- ci has logical network as IP/24, the lport number must < 256
- lport_bind operation does not support batch size > 1 now in
  container-based deployment
- Reduce the lport in the ci job as we do not run real scale test
  on travis-ci VMs

### DIFF
--- a/ci/ansible/all.yml
+++ b/ci/ansible/all.yml
@@ -36,8 +36,8 @@ ovn_number_chassis: 5
 ########################
 network_start_cidr: "172.16.201.0/24"
 network_number: "5"
-ports_per_network: "500"
-ports_created_batch_size: "50"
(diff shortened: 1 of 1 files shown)


ci/ansible/all.yml
```diff
@@ -36,8 +36,8 @@ ovn_number_chassis: 5
 ########################
 network_start_cidr: "172.16.201.0/24"
```
mestery, line 37: I think we should change this CIDR instead.

mestery:
I think it's fine to change this for the upstream CI for now, though I would ideally like more than 10 ports. We'll modify our downstream CI to use 500 ports instead.

PR #100, opened 2016-08-12
Unset 40th bit of the lport address (broadcast)

huikang:
Thanks, Ryan.


mestery:
Looks like the build failed.

PR #101, opened 2016-08-12
Fix gate docker version

huikang:


huikang:
the error message was https://travis-ci.org/openvswitch/ovn-scale-test#L262

Issue #97, opened 2016-07-28
Use ovn-nbctl --wait hv to get end-to-end port binding time


hzhou8:
Currently the scale testing waits port state UP in ovn-nb and then starts next round of port creation and binding. This is inaccurate, because when CPU is 100% in test farms it can't reflect the real time that ovn-controllers spent to complete each round of processing. 

With the new feature "--wait hv" [1], we can now wait for the port bindings to be processed and reflected on all HVs before starting next round, so that the real processing time can be reflected and get more accurate result of port-binding performance.

[1] https://github.com/openvswitch/ovs/commit/fa183acc654f7e5da17cd70fc91d6b5b02782183

huikang:
@hzhou8 would you mind being assigned of this task? Thanks.

huikang:
@hzhou8 the commit message of this feature says
`With --wait=sb, ovn-nbctl now waits for ovn-northd
to update the southbound database; with --wait=hv, it waits for the
changes to make their way to Open vSwitch on every hypervisor.`

However, ovn-scale-test uses "ovs-vsctl" command to bind the logical port
to br-int on the hypervisor, instead of ovn-nbctl. Could you clarify that at which step, shall we
add "--wait hv" in ovn-scale-test? Thanks.

hzhou8:
@huikang yes this is a little bit tricky because port binding with ovs-vsctl happens on HV, which doesn't change ovn-nb directly. Instead, it is ovn-controller detect the port-binding and then update the ovn-sb, which is then detected by ovn-northd which updates the port state in ovn-nb. 

So instead of making a change by ovn-nbctl and waiting for the HVs to catch up, we use wait-until to wait for port state become up (this behavior doesn't change), and then use "ovn-nbctl --wait=hv sync" to wait for the port state change to be processed on all HVs.

I submitted a pull request [1] for this. Please kindly review. Thanks!

[1] https://github.com/openvswitch/ovn-scale-test/pull/120

huikang:
@hzhou8 can we close this issue now as the PR has been merged? Thanks.

hzhou8:
Yes :)

PR #102, opened 2016-08-14
Check openflow(OF) rules on each hyperviors after binding lports

huikang:
- use ovs-appctl to trace flow matches in the tables

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -15,10 +15,12 @@
 import pipes
 import sys
 import itertools
-import pipes
+import StringIO
+from rally.common import logging
 from rally_ovs.plugins.ovs.ovsclients import *
 from rally_ovs.plugins.ovs.utils import get_ssh_from_credential
 
+LOG = logging.getLogger(__name__)
 
 @configure("ssh")
 class SshClient(OvsClient):
@@ -167,7 +169,7 @@ class OvnNbctl(OvsClient):
 
         def show(self, lswitch=None):
             params = [lswitch] if lswitch else []
-            stdout = StringIO()
+            stdout = StringIO.StringIO()
             self.run("show", args=params, stdout=stdout)
             output = stdout.getvalue()
 
@@ -203,7 +205,7 @@ class OvsVsctl(OvsClient):
             self.sandbox = sandbox
             self.install_method = install_method
 
-        def run(self, cmd, opts=[], args=[]):
+        def run(self, ovs_cmd, cmd, opts=[], args=[], stdout=sys.stdout):
             self.cmds = self.cmds or []
 
             # TODO: tested with non batch_mode only for docker
@@ -213,16 +215,17 @@ class OvsVsctl(OvsClient):
             if self.sandbox and self.batch_mode == False:
                 if self.install_method == "sandbox":
                     self.cmds.append(". %s/sandbox.rc" % self.sandbox)
-                    cmd = itertools.chain(["ovs-vsctl"], opts, [cmd], args)
+                    cmd = itertools.chain([ovs_cmd], opts, [cmd], args)
                     self.cmds.append(" ".join(cmd))
                 elif self.install_method == "docker":
-                    self.cmds.append("sudo docker exec %s ovs-vsctl " % self.sandbox + cmd + " " + " ".join(args))
+                    self.cmds.append("sudo docker exec %s " % self.sandbox + " "
+                                     + ovs_cmd + " " + cmd + " " + " ".join(args))
 
             if self.batch_mode:
                 return
 
             self.ssh.run("\n".join(self.cmds),
-                         stdout=sys.stdout, stderr=sys.stderr)
+                         stdout=stdout, stderr=sys.stderr)
 
             self.cmds = None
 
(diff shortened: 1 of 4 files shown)


rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -213,16 +215,16 @@ def run(self, cmd, opts=[], args=[]):
             if self.sandbox and self.batch_mode == False:
                 if self.install_method == "sandbox":
                     self.cmds.append(". %s/sandbox.rc" % self.sandbox)
-                    cmd = itertools.chain(["ovs-vsctl"], opts, [cmd], args)
+                    cmd = itertools.chain([ovs_cmd], opts, [cmd], args)
                     self.cmds.append(" ".join(cmd))
                 elif self.install_method == "docker":
-                    self.cmds.append("sudo docker exec %s ovs-vsctl " % self.sandbox + cmd + " " + " ".join(args))
+                    self.cmds.append("sudo docker exec %s " % self.sandbox + ovs_cmd + cmd + " " + " ".join(args))
```
hzhou8: add space here: ... self.sandbox + " " + ovs_cmd + " " ...

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -242,13 +244,36 @@ def flush(self):
 
         def add_port(self, bridge, port, may_exist=True):
             opts = ['--may-exist'] if may_exist else None
-            self.run('add-port', opts, [bridge, port])
+            self.run(" ovs-vsctl ", 'add-port', opts, [bridge, port])
```
hzhou8: remove the space here: "ovs-vsctl"

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -242,13 +244,36 @@ def flush(self):
...
 
 
         def db_set(self, table, record, *col_values):
             args = [table, record]
             args += set_colval_args(*col_values)
-            self.run("set", args=args)
+            self.run(" ovs-vsctl ", "set", args=args)
```
hzhou8: same as above

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -242,13 +244,36 @@ def flush(self):
...
+
+        def of_check(self, bridge, port, mac_addr, may_exist=True):
+            in_port = ""
+            stdout = StringIO.StringIO()
+            self.run(" ovs-ofctl ", "show br-int", stdout=stdout)
```
hzhou8: same as above

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -242,13 +244,36 @@ def flush(self):
...
+            show_br_int_output = stdout.getvalue()
+            in_port = get_of_in_port(show_br_int_output, port)
+            LOG.info("Check port: in_port: %s; mac: %s" % (in_port.strip(), mac_addr))
+            appctl_cmd = " ofproto/trace br-int in_port=" + in_port.strip()
```
hzhou8, line 263: If the purpose is to verify the binding operation finished and reflected in flows, how about simply: ofctl dump-flows br-int table=<the arp table>| grep <mac address>

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -251,6 +254,43 @@ def _create_networks(self, network_create_args):
         return lswitches
 
 
+    @atomic.action_timer("ovn_network.of_check_port")
+    def _of_check_ports(self, lports, sandboxes, port_bind_args):
```
hzhou8, line 258: Could you write some comments about why, what and how is it checked?

rally_ovs/plugins/ovs/ovsclients_impl.py
huikang, line 263: @hzhou8 thanks for reviewing the PR. You are right; binding operation could be verified by `ofctl dump-flow`. However, IMO, tracing a flow, which is used [OVN tutorial](https://github.com/openvswitch/ovs/blob/master/tutorial/ovn/env3/packet1.sh#L19), seems more accurate and straightforward.

rally_ovs/plugins/ovs/scenarios/ovn.py
huikang, line 258: **Why?** In our OVN scalability test with dataplane (NOTE: not ovn-scale-test/sandbox), we found that some lport is not pingable after binding. After certain time period, the lport becomes pingable. One suspicious point is that the openflow rules after binding is not setup fast enough. Therefore checking "port state up" is insufficient.
**What?** In ovn-scale-test, since there is no real kernel module installed, the ping tool is not available. Therefore, we have to use OVS trace tool to verify the openflow rules.
**How?** Use the OVS built-in trace tool to verify the openflow rules are set correctly.

rally_ovs/plugins/ovs/ovsclients_impl.py
hzhou8, line 263: Using the flow trace seems too heavy and subject to change in the future. For example, when there are ACL rules, the programmed flows can also result in a drop action. So I think dump-flow for the ARP table is straightforward and simple. And less operations, e.g. no need to find out the ofport first.

rally_ovs/plugins/ovs/ovsclients_impl.py
huikang: done

rally_ovs/plugins/ovs/ovsclients_impl.py
huikang: done

rally_ovs/plugins/ovs/ovsclients_impl.py
huikang: done

rally_ovs/plugins/ovs/ovsclients_impl.py
huikang: done

rally_ovs/plugins/ovs/scenarios/ovn.py
hzhou8, line 258: If I understand correctly, the purpose is to evaluate the end-to-end time spent in port-binding and flow installation in HVs, right? Yes, checking "port state up" is insufficient. But there are problems of checking flows on HVs:
1. It is not accurate. The periodical poll have gaps.
2. We need to check all _related_ HVs (sandboxes), instead of only the ones doing port-binding. And because of this, this approach doesn't scale, because checking flows on so many sandboxes and collect the data would take time.

Because of this, we need a notification mechanism to accurately evaluate when all the HVs have finished processing the port-binding. It is handled by [1]. I believe it is a better approach, but I haven't worked on it yet.

[1] https://github.com/openvswitch/ovn-scale-test/issues/97

huikang, line 258: @hzhou8 I agreed with both of your points; we should use the `--wait hv` [1] to measure the latency of binding lport. But I still have two questions:
1. In #97, can we simply add `--wait hv` to all the nb-ctl command so that the time in rally-ovs results reflect the binding time from issuing the command to the OF rules installed on all involved hypervisors? Is this enough?
2. Even with `--wait hv`, we somehow still want verify the correctness of the flows. As flow tracing is too heave as your mentioned, do you think I can implement it as a asynchronous thread? In other words, besides the main rally-ovs thread, a separate python thread is launched to trace the flow. What do you think? Thanks.

[1] #97

hzhou8, line 258: @huikang 
For question 1): Hopefully that simple change would work, but I haven't tried myself yet. I guess I won't have time this week, so it would be highly appreciated if you could try. Let me know if you'd like to take the ticket :)

For 2): I wonder if it is really needed to check correctness of flows in scale test. IMHO the focus here is the performance, and OVN functional tests should ensure correctness (the tests under OVS repo, and also openstack networking-ovn etc.). If we do suspect some scalability related bug, it may be helpful to have such kind of check AFTER the scale test complete. If we decide to implement such checks, we should make sure those checks won't impact the performance data. Instead of a separate thread running in parallel, I'd prefer a post-test validation. It should not be enabled by default because it would take too long considering the number of ports in scale test, and only enabled when we are testing correctness. Priority should not be high, though. What do you think?

huikang, line 258: @hzhou8 Thanks for your response.
1. I am very happy to give the first shot and will try to submit a PR to #97 soon.
2. Putting flow verification as an option in the post-test stage sounds good to me.

PR #104, opened 2016-08-15
Initialize north and south db

huikang:
Particularly initialize NB_Global and SB_Global tables, whcih is
used by --wait of ovn-nbctl.


huikang:
These two commands has been used in the ovn-sandbox script [1]

[1] https://github.com/openvswitch/ovs/blob/master/tutorial/ovs-sandbox#L364

PR #105, opened 2016-08-19
Run ovn-scale-test without rally

mestery:
Add the capability to run this without Rally in preparation for adding pure
OVN tests here.

### DIFF
--- a/ci/scale-pure-ovn-hosts.sh
+++ b/ci/scale-pure-ovn-hosts.sh
@@ -23,7 +23,7 @@ $OVNSUDO docker images
 
 # Deploy the containers
 pushd $OVN_SCALE_TOP
-$OVNSUDO /usr/local/bin/ansible-playbook -i $OVN_DOCKER_HOSTS ansible/site.yml -e @$OVN_DOCKER_VARS \
+$OVNSUDO /usr/local/bin/ansible-playbook -i $OVN_DOCKER_HOSTS ansible/site-ovn-only.yml -e @$OVN_DOCKER_VARS  -e enable_rally_ovs="no" \
      --extra-vars "ovs_repo=$OVS_REPO" --extra-vars "ovs_branch=$OVS_BRANCH" --extra-vars "configflags=$CONFIG_FLAGS" -e action=deploy
 if [ "$?" != "0" ] ; then
     echo "Deploying failed, exiting"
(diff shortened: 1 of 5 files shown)


huikang:
I like this change. However, I preferring to have one site.yml for easy management. We can use a variable like "enable_rally_ovs". Then in existing site.yml,  if "when: enable_rally_svs | bool", deploy the rally container.

huikang:
For the pure OVN test, we can add a second travis-ci job.

ci/scale-pure-ovn-hosts.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Save trace setting
+XTRACE=$(set +o | grep xtrace)
+set -o xtrace
+
+# Read variables
+source ovn-scale.conf
```
huikang, line 8: Shall we add the above to `docker-rally.sh`?

mestery, line 8: Eventually, but now it doesn't run any tests, so lets wait until I do that. This is the first step. :)

ci/scale-pure-ovn-hosts.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# Library files
+source scale-lib.sh
```
huikang, line 11: @mestery sorry, I put the comment below the wrong line. Since the functions (e.g., `check_container_failure`) are moved to scale-lib.sh, we should source the file in original driver file, which is docker-rally.sh

mestery, line 11: We could, but I run these individually sometimes, so I'd rather leave it in each file, similar to line 8 where it sources the .conf file as well.

huikang, line 11: Ok, understood. Wait for the ci now. Thanks.

huikang:
@mestery the CI test has passed and this PR should be merged. But since you are on this ci script, do you mind fixing [this issue](https://travis-ci.org/openvswitch/ovn-scale-test/builds/153662512#L6423-L6427) as well? The problem is caused by this commented [line](https://github.com/openvswitch/ovn-scale-test/blob/master/ci/scale-test.sh#L30). Since the task UUID is not deleted, the next task returns two UUIDs, which confuses the `rally task report` command.

But if you think the above issue should be tackled by another PR, I am ok to merge this one. Thanks.

mestery:
@huikang yes, I will push a new version addressing the issue you note by uncommenting the line above. Thanks!

mestery:
@huikang New version pushed addressing your comment.

huikang:
@mestery thanks for updating the ci script. There is a duplicate function definition of `function check_on_rally`. Since you moved it to scale-lib.sh, could you remove the same function at [here](https://github.com/mestery/ovn-scale-test/blob/pure-ovn/ci/scale-test.sh#L10)?

mestery:
@huikang Done!

mestery:
@huikang This one is now ready I believe.

huikang:
@mestery thanks!

PR #109, opened 2016-08-24
Enhance ci/prepare.sh when appending to /etc/ssh/sshd_config

flavio-fernandes:
The file /etc/ssh/sshd_config -- from official trusty vagrant box --
is not terminated with the end-of-line character.

By appending to that file, the "UsePAM yes" config gets corrupted
and subsequent ssh access starts to fail.

With this change -- if needed -- a new line will be added to that
file before any potential append takes place.

   # diff orig/ssh/sshd_config /etc/ssh/sshd_config
   88c88,91
   < UsePAM yes
   \ No newline at end of file

---

> UsePAM yesMatch host 10.0.2.15
>     PermitRootLogin without-password

Signed-off-by: Flavio Fernandes flavio@flaviof.com


huikang:
LGTM; but I would like to wait for @mestery 's feedback. Thanks.

mestery:
LGTM, thanks Flavio!

PR #112, opened 2016-08-28
prepare.sh: Create ${OVN_SCALE_TOP}/ansible/site-ovn-only.yml

flavio-fernandes:
File site-ovn-only.yml is created from site.yml.

Nitpick: remove extra space character, when creating
/etc/apt/sources.list.d/docker.list


flavio-fernandes:
To be asked:
- maybe add site-ovn-only.yml to .gitignore?
- maybe use a real yaml parser and extract the wanted section out of site.yml?

flavio-fernandes:
NOTE: this change is actually a complement of [pull request 105](https://github.com/openvswitch/ovn-scale-test/pull/105)

huikang:
The ovn-only-site.yml should be specific to deployment with no rally_ovs. Therefore, we will have two travis job: one uses rally_ovs and one not. Since the non-rally travis job does not exist yet, I suggest hold this PR until the travis job is added. Thanks.

PR #113, opened 2016-08-29
Fix missing batch size support in create_network

huikang:
Without batch size support, all ovn-set command will be issued in
one single command, which may exceed the limit of OVN (e.g., > 10k
networks in a single command line).

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -231,6 +231,7 @@ class OvnScenario(scenario.OvsScenario):
     def _create_networks(self, network_create_args):
         physnet = network_create_args.get("physical_network", None)
         lswitches = self._create_lswitches(network_create_args)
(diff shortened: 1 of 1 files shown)


huikang:
Hi, @l8huang , could you take a look at this PR? Thanks.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -231,13 +231,15 @@ def _delete_acl(self, lswitches):
     def _create_networks(self, network_create_args):
         physnet = network_create_args.get("physical_network", None)
         lswitches = self._create_lswitches(network_create_args)
+        batch = network_create_args.get("batch", 1)
```
l8huang: suggest to use the number of lswitches as default value:
batch = network_create_args.get("batch", len(lswitches))

l8huang:
one small comments :)

mestery:
This LGTM after addressing the comment from @l8huang, thanks!

PR #115, opened 2016-08-29
prepare.sh: Move creation of ansible/docker-ovn-hosts

flavio-fernandes:
And avoid duplicate code in determining local ip

### DIFF
--- a/ci/prepare.sh
+++ b/ci/prepare.sh
@@ -65,14 +65,6 @@ sudo apt-get remove -y ansible
 sudo pip install ansible==2.0.2.0
 sudo pip install --upgrade setuptools
 
-# Prepate the docker-ovn-hosts file
-LOCALIP=$(ip addr show dev eth0 | grep 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
-if [ "$LOCALIP" == "" ] ; then
-    # Try bond0
-    LOCALIP=$(ip addr show dev bond0 | grep 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
-fi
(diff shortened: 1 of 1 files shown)


ci/prepare.sh
```diff
@@ -65,14 +65,6 @@ sudo apt-get remove -y ansible
 sudo pip install ansible==2.0.2.0
 sudo pip install --upgrade setuptools
 
-# Prepate the docker-ovn-hosts file
-LOCALIP=$(ip addr show dev eth0 | grep 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
-if [ "$LOCALIP" == "" ] ; then
-    # Try bond0
-    LOCALIP=$(ip addr show dev bond0 | grep 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
-fi
-cat ansible/docker-ovn-hosts-example | sed -e "s/REPLACE_IP/$LOCALIP/g" > ansible/docker-ovn-hosts
```
huikang, line 74: With these line removed, how can the task discover the IP of the host in the ci job? Thanks.

flavio-fernandes: @huikang: the lines removed are actually duplicated down below in that same file.
With that, they are just not needed.
Note that line 74 is moved, not removed, to the lines below.

huikang: @flavio-fernandes thanks for your explanation. LGTM. 

@mestery could you review this as well? Thanks.

mestery: @flavio-fernandes Actually, line 74 is not duplicated. I think I'd rather you move some of the new logic down below (around line 117) upwards here as we need to set LOCALIP in ci/ansible/docker-ovn-hosts. Make sense?

mestery: NM, I see this now. :)

PR #118, opened 2016-08-31
Handle getting IP information for interfaces with multiple addresses

mestery:
PR #115 [1] moved where we setup the ansible/docker-ovn-hosts file
to later in the script. This commit moves this back to the top of the
file to increase readability.

Further, this commit fixes a serious issue for interfaces with multiple
addresses. Currently, if a system has many addresses per interface (see
below), the current logic will collect all the IP addresses, which will
fail in setting up ansible/docker-ovn-hosts. This adds "-m 1" to the grep
command to only get the first match. Well less than ideal, it works and
allows things to work.

This was failing in the case like below where bond0 has many IP
addresses:

root@redacted:~# ip addr show dev bond0
6: bond0: <BROADCAST,MULTICAST,MASTER,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 0c:c4:7a:39:9a:38 brd ff:ff:ff:ff:ff:ff
    inet 10.137.110.167/26 brd 10.137.110.191 scope global bond0
       valid_lft forever preferred_lft forever
    inet 172.16.20.100/16 scope global bond0
       valid_lft forever preferred_lft forever
    inet 172.16.200.10/16 scope global secondary bond0
       valid_lft forever preferred_lft forever
    inet 172.16.200.11/16 scope global secondary bond0
       valid_lft forever preferred_lft forever
    inet 172.16.200.12/16 scope global secondary bond0
       valid_lft forever preferred_lft forever
    inet 172.16.200.13/16 scope global secondary bond0
       valid_lft forever preferred_lft forever
    inet 172.16.200.14/16 scope global secondary bond0
       valid_lft forever preferred_lft forever
    inet6 fe80::ec4:7aff:fe39:9a38/64 scope link
       valid_lft forever preferred_lft forever
root@redacted:~#

[1] https://github.com/openvswitch/ovn-scale-test/pull/115

### DIFF
--- a/ci/prepare.sh
+++ b/ci/prepare.sh
@@ -31,6 +31,22 @@ if [ "$INSTALLDOCKER" == "True" ] ; then
     fi
 fi
 
+# Determine local ip of this system.
+# First, try eth0. Then, try bond0. Lastly, use the address
+# of the interface that is used by the default route.
+LOCALIP=$(ip addr show dev eth0 | grep -m 1 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
+if [ "$LOCALIP" == "" ] ; then
+    # Try bond0
+    LOCALIP=$(ip addr show dev bond0 | grep -m 1 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
+fi
+if [ "$LOCALIP" == "" ] ; then
+    # Try to use interface used in default route
+    PHYS_DEV=$(ip route list match 0.0.0.0/0 | grep -m 1 -oP "(?<=dev )[^\s]*(?=\s)")
+    LOCALIP=$(ip -4 addr show $PHYS_DEV | grep -m 1 -oP "(?<=inet ).*(?=/)")
+fi
+# Prepare the docker-ovn-hosts file
+cat ansible/docker-ovn-hosts-example | sed -e "s/REPLACE_IP/$LOCALIP/g" > ansible/docker-ovn-hosts
+
 # Setup ssh
 if [ ! -f ~/.ssh/id_rsa.pub ] ; then
     ssh-keygen -t rsa -f ~/.ssh/id_rsa -N ''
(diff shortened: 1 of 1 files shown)


mestery:
@flavio-fernandes Have a peek here when you have time, thanks!

flavio-fernandes:
will do, added to my todo.

mestery:
This is weird. travis-ci failed here because in prepare.sh, the parts where we're calling "$OVNSUDO su -m root" are not being run [1] [2] [3] [4]. See here [5] for the failure.

Looking at a PR from @huikang (PR #113), you can see they also do not run there [6]. More investigation needed.

[1] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/prepare.sh#L47
[2] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/prepare.sh#L54
[3] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/prepare.sh#L62
[4] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/prepare.sh#L87
[5] https://travis-ci.org/openvswitch/ovn-scale-test/builds/156406269#L517
[6] https://travis-ci.org/openvswitch/ovn-scale-test/builds/156223348#L508

ci/prepare.sh
```diff
@@ -7,6 +7,22 @@ set -o xtrace
...
+# First, try eth0. Then, try bond0. Lastly, use the address
+# of the interface that is used by the default route.
+LOCALIP=$(ip addr show dev eth0 | grep -m 1 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
+if [ "$LOCALIP" == "" ] ; then
+    # Try bond0
+    LOCALIP=$(ip addr show dev bond0 | grep -m 1 'inet ' | cut -d " " -f 6 | cut -d "/" -f 1)
+fi
+if [ "$LOCALIP" == "" ] ; then
+    # Try to use interface used in default route
+    PHYS_DEV=$(ip route list match 0.0.0.0/0 | grep -oP "(?<=dev )[^\s]*(?=\s)")
```
flavio-fernandes: let's put the -m1 here too!
ip route list match 0.0.0.0/0 | grep -m 1 -oP "(?<=dev )[^\s]*(?=\s)"

flavio-fernandes:
LGTM!

As we discussed, sections of the script inside <<EOF delimiters are not shown in the trace, but they do get executed and all we see is the output of any command(s) that are inside.

huikang:
we need to dump out the content of /home/travis/build/openvswitch/ovn-scale-test/ci/ansible/docker-ovn-hosts, as the next step failed  in travis-ci [1]

[1] https://travis-ci.org/openvswitch/ovn-scale-test/builds/156641038#L7617

mestery:
@huikang I will push a new version that dumps the file now.

huikang:
@mestery It seems an ssh error. With this patch, the ssh configuration happens after preparing the ovn-host inventory file [1].

To debug, can we add something `ssh $LOCALIP ls` to see if ssh works?

[1] https://github.com/openvswitch/ovn-scale-test/blob/master/ci/prepare.sh#L34-L51

flavio-fernandes:
I think I got a better idea on what is going on here.
When the script does "$OVNSUDO su -m root <<'EOF'" in line 104
the variable LOCALIP is no longer scoped. Because of that, your sshd
config file will look like this:

tail /etc/ssh/sshd_config
...
UsePAM yes
Match host         <= bleh!
    PermitRootLogin without-password
Match host 172.17._._
    PermitRootLogin without-password

And that makes a very #sadpanda happen.

mestery:
That was it @flavio-fernandes! New version inbound.

huikang:
Thanks for all your works. The PR can be merged now.

PR #120, opened 2016-09-01
Use ovn-nbctl --wait=hv sync to get end-to-end port binding time

hzhou8:
Currently the scale testing waits port state UP in ovn-nb and then
starts next round of port creation and binding. This is inaccurate,
because when CPU is 100% in test farms it can't reflect the real
time that ovn-controllers spent to complete each round of processing.

With the new feature "--wait=hv" [1], and the follow-up "sync" cmd
support [2], we can now wait for the port bindings to be processed
and reflected on all HVs before starting next round, so that the real
processing time can be reflected and evaluate more accurately the
related scalability improvements.

[1] https://github.com/openvswitch/ovs/commit/fa183acc654f7e5da17cd70fc91d6b5b02782183
[2] https://github.com/openvswitch/ovs/commit/de32cec780e8d2a4358075c2449858247881f266

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -263,6 +263,13 @@ class OvnScenario(scenario.OvsScenario):
     def _bind_ports(self, lports, sandboxes, port_bind_args):
         port_bind_args = port_bind_args or {}
         wait_up = port_bind_args.get("wait_up", False)
+        # "wait_sync" takes effect only if wait_up is True.
+        # By default we wait for all HVs catching up with the change.
+        wait_sync = port_bind_args.get("wait_sync", "hv")
+        if wait_sync.lower() not in ['hv', 'sb', 'none']:
+            raise exceptions.InvalidConfigException(_(
+                "Unknown value for wait_sync: %s. "
+                "Only 'hv', 'sb' and 'none' are allowed.") % wait_sync)
 
         sandbox_num = len(sandboxes)
         lport_num = len(lports)
@@ -294,12 +301,12 @@ class OvnScenario(scenario.OvsScenario):
             j += 1
 
         if wait_up:
-            self._wait_up_port(lports, install_method)
+            self._wait_up_port(lports, wait_sync, install_method)
 
 
     @atomic.action_timer("ovn_network.wait_port_up")
-    def _wait_up_port(self, lports, install_method):
-        LOG.info("wait port up" )
+    def _wait_up_port(self, lports, wait_sync, install_method):
+        LOG.info("wait port up. sync: %s" % wait_sync)
         ovn_nbctl = self.controller_client("ovn-nbctl")
         ovn_nbctl.set_sandbox("controller-sandbox", install_method)
         ovn_nbctl.enable_batch_mode(True)
@@ -307,8 +314,8 @@ class OvnScenario(scenario.OvsScenario):
         for lport in lports:
             ovn_nbctl.wait_until('Logical_Switch_Port', lport["name"], ('up', 'true'))
 
-        ovn_nbctl.flush()
-
+        if wait_sync != "none":
+            ovn_nbctl.sync(wait_sync)
 
 
 
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -175,6 +175,17 @@ class OvnNbctl(OvsClient):
 
             return get_lswitch_info(output)
 
+        def sync(self, wait='hv'):
+            # sync command should always be flushed
+            opts = ["--wait=%s" % wait]
+            batch_mode = self.batch_mode
+            if batch_mode:
+                self.flush()
+                self.batch_mode = False
+
+            self.run("sync", opts)
+            self.batch_mode = batch_mode
+
     def create_client(self):
         print "*********   call OvnNbctl.create_client"
 
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -57,12 +57,14 @@ class OvnNbctl(OvsClient):
                 return
 
             if self.sandbox:
+                cmd_prefix = []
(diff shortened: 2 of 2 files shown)


huikang:
@hzhou8 will do. Thanks.

huikang:
@hzhou8 the changes to rally_ovs are good. However, I think this `--wait` should be configurable in the rally_task file. For example, some options to the variable could be `NONE, sb, or hv`.

hzhou8:
@huikang thanks for the review!
Maybe I didn't get your point. For sync command, there are only 2 choices: wait=sb or wait=hv, because sync without --wait is meaningless. The function OvnNbctl.sync() has parameter to tell if it is sb or hv it wants to wait, and default is hv.

For port-binding, we want to wait for hv to be synced, so use the default parameter.

Do you mean a general configuration of --wait for all the commands, or something else?

huikang:
@hzhou8 sorry about the confusion. My thought was to add something like `wait_sync:` in the general rally_ovs task file [1]. But we can handle this in a separate PR if you prefer. Thanks.

[1] https://github.com/openvswitch/ovn-scale-test/blob/master/ansible/roles/rally/templates/create_networks.json.j2

hzhou8:
@huikang ok, thanks for explain. I thought we either wait for port up and synced, or we don't wait, which is already configurable by "wait_up". But yes it may be valuable to be able to wait port up in north-bound only. I updated the patch according to your suggestion.

huikang:
LGTM. Could @mestery or @l8huang review this one also?
Thanks.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -263,6 +263,13 @@ def _create_networks(self, network_create_args):
     def _bind_ports(self, lports, sandboxes, port_bind_args):
         port_bind_args = port_bind_args or {}
         wait_up = port_bind_args.get("wait_up", False)
+        # "wait_sync" takes effect only if wait_up is True.
+        # By default we wait for all HVs catching up with the change.
+        wait_sync = port_bind_args.get("wait_sync", "hv")
+        if wait_sync not in ['hv', 'sb', 'none']:
```
l8huang: better to convert wait_sync to lower case

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -263,6 +263,13 @@ def _create_networks(self, network_create_args):
...
+            raise exceptions.InvalidConfigException(_(
```
l8huang, line 270: Shall ovs version being checked, or leave a comment here to state this option is available since which verion.

l8huang:
2 Comments added. Most part LGTM, please merge it once you think it's good to merge.

rally_ovs/plugins/ovs/scenarios/ovn.py
huikang: @l8huang I think ovs version check should be executed in `ovsclients_impl.py`.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -263,6 +263,13 @@ def _create_networks(self, network_create_args):
     def _bind_ports(self, lports, sandboxes, port_bind_args):
         port_bind_args = port_bind_args or {}
         wait_up = port_bind_args.get("wait_up", False)
+        # "wait_sync" takes effect only if wait_up is True.
```
huikang, line 266: @hzhou8 Shall we add some condition check for `wait_up is True` as the comment says?

rally_ovs/plugins/ovs/scenarios/ovn.py
hzhou8: Currently ovn is not officially released yet (first release will be in ovs 2.6) and the scale-test is always based on master, so no need to check version.

rally_ovs/plugins/ovs/scenarios/ovn.py
hzhou8: Done

rally_ovs/plugins/ovs/scenarios/ovn.py
hzhou8: The code already does as the comment says: if wait_up is False the wait_sync won't take effect.

huikang: @hzhou8 sorry for missing the condition check code below. If so, shall we move the above comment near `if wait_up`?

hzhou8: @huikang this comment is to help understand the usage of this parameter without having to read through the code, hence it is put here (like definition). Otherwise, the comment may be removed.

huikang:
@hzhou8 thanks for your explanation; I agree keeping this code as it is.

The travis-ci job somehow failed [1] due to some OVS version issue, while our ci script couldn't report this [2]. @mestery , could you take a look at this?. 

[1] https://travis-ci.org/openvswitch/ovn-scale-test/builds/157977687#L7322 
[2] https://travis-ci.org/openvswitch/ovn-scale-test/builds/157977687#L7450

huikang:
@mestery I submitted a patch to fix the false positive error. Could you review #124? We will re-run the travis-ci job after #124 being merged. Thanks.

hzhou8:
@huikang thanks for the catch! It is not because of ovs version but because of the batch mode caused the extra "--" added before --wait, which is a global option instead of a sub-command option, and that's why "ovn-nbctl -- --wait=hv sync" will report error saying that 'sync' command has no '--wait' option, while "ovn-nbctl --wait=hv sync" works. Here is the fix to avoid that "--" by not using batch_mode.

While working on the fix, I noticed another bug: when using docker the opts were not passed to the cmd line if not using batch_mode. Here is the fix for this: https://github.com/openvswitch/ovn-scale-test/pull/125

hzhou8:
Just checked the CI logs and now the checks are _really_ passed, together with patch #125

huikang:
@hzhou8 Thanks for the fix. I will pull your branch and verify the results in my local environment; once it passes, I will merge the PR.

PR #123, opened 2016-09-07
Fix warning: use file module with owner instead of chown

flavio-fernandes:
Replace command with file modules, to address the following warning
in ansible/roles/ovn/tasks/config.yml:

 [WARNING]: Consider using file module with owner rather than running chown


huikang:
Thanks.

PR #125, opened 2016-09-07
Fix bug: opts are missing for docker cmd execution

hzhou8:


huikang:
@hzhou8 I introduced this bug when implementing the container-based deployment. Let me take a closer look at this. Thanks @hzhou8 !

mestery:
This LGTM, @huikang can we merge this one?

huikang:
sorry for the late response

PR #129, opened 2016-09-12
ci/scale-pure-ovn-hosts: remove unused params

flavio-fernandes:
The params: ovs_repo, ovs_branch, configflags are not needed
when running the deploy, as these are only used in the construction
of the docker images.

### DIFF
--- a/ci/scale-pure-ovn-hosts.sh
+++ b/ci/scale-pure-ovn-hosts.sh
@@ -22,9 +22,14 @@ popd
 $OVNSUDO docker images
 
 # Deploy the containers
+# TODO(flaviof): When deploy is hooked into a CI system to test a specific OVS commit sha,
+# we will need to add support for passing in the OVS_REPO, OVS_BRANCH, and CONFIG_FLAGS
+# variables. Till then, let's keep it out to avoid potential confusion.
 pushd $OVN_SCALE_TOP
-$OVNSUDO /usr/local/bin/ansible-playbook -i $OVN_DOCKER_HOSTS ansible/site-ovn-only.yml -e @$OVN_DOCKER_VARS  -e enable_rally_ovs="no" \
-     --extra-vars "ovs_repo=$OVS_REPO" --extra-vars "ovs_branch=$OVS_BRANCH" --extra-vars "configflags=$CONFIG_FLAGS" -e action=deploy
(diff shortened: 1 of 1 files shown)


huikang:
These variables are used in several places in creating rally_ovs task file. However, if pure-ovn deployment does not depends on rally, removing these variables looks fine.

mestery:
I think we will want these if we ever use "pure-ovn" in a CI system which is testing patches, unless I'm missing something.

flavio-fernandes:
yeah, I can see the potential for using them in the future, but right now I
find it is a bit confusing and redundant. I'm okay with abandoning this PR
if you think it is still better to have it as part of the deploy. No harm. ;)

huikang:
@flavio-fernandes we can keep this PR for some time to see if we really need or can improve it. Thanks.

ci/scale-pure-ovn-hosts.sh
```diff
@@ -23,8 +23,10 @@ $OVNSUDO docker images
 
 # Deploy the containers
 pushd $OVN_SCALE_TOP
-$OVNSUDO /usr/local/bin/ansible-playbook -i $OVN_DOCKER_HOSTS ansible/site-ovn-only.yml -e @$OVN_DOCKER_VARS  -e enable_rally_ovs="no" \
-     --extra-vars "ovs_repo=$OVS_REPO" --extra-vars "ovs_branch=$OVS_BRANCH" --extra-vars "configflags=$CONFIG_FLAGS" -e action=deploy
+$OVNSUDO /usr/local/bin/ansible-playbook -i $OVN_DOCKER_HOSTS ansible/site-ovn-only.yml \
```
mestery, line 29: How about adding a comment above this (something like this):

# TODO(flaviof): When this is hooked into a CI system to test a specific OVS commit sha, we will need

# to add back support for passing in the OVS_REPO, OVS_BRANCH, and CONFIG_FLAGS variables.

mestery:
I am fine merging this once a comment is added @flavio-fernandes, thanks for the work here!

ci/scale-pure-ovn-hosts.sh
flavio-fernandes: sounds like a very reasonable compromise. Thanks Kyle!

PR #130, opened 2016-09-16
Add batch processing variable for network creation

mestery:
Without this, you cannot create more than 500 networks or so, as we exceed
the command line length.


mestery:
@huikang Please review this.

huikang:
Thanks for fixing this.

PR #135, opened 2018-01-05
Fix rally-ovs repo path in install.sh

noah8713:
Fix the typo in the rally-ovs repo which is merged as rally in upstream to help avoid error while setting up ovn scale test environment.


hzhou8:
Thanks for the patch! Could you update the comments a little bit to avoid confusion? I think nothing is merged to rally in upstream. It is just a typo. Also, please update your email since it is not recognized :)

noah8713:
@hzhou8  Sure. Sorry for the confusion.  Following the steps at http://ovn-scale-test.readthedocs.io/en/latest/install.html , l8huang/rally repo already have the base patch used to do this scale test with rally. However, the same in install.sh is wrong. Hence when setting up rally, its better to use the same code base to avoid extra re-work if something is wrong in integration.

Email might not be visible due to the settings. Its visible now. :)

PR #137, opened 2018-01-26
Add overlay test cases

noah8713:


rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -247,7 +247,7 @@ def run(self, cmd, opts=[], args=[]):
                 elif self.install_method == "docker":
                     self.cmds.append("sudo docker exec %s ovs-vsctl " % self.sandbox + cmd + " " + " ".join(args))
 
-            if self.install_method != "docker"
+            if self.install_method != "docker":
```
hzhou8: Could you create a separate patch for this fix? Thx!

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -377,6 +377,39 @@ def _bind_ports(self, lports, sandboxes, port_bind_args):
         if wait_up:
             self._wait_up_port(lports, wait_sync, install_method)
 
+    @atomic.action_timer("ovn_network.bind_overlay_ports")
+    def _bind_overlay_ports(self, farms, port_bind_args):
```
hzhou8: this one seems redundant with _bind_ports. It is better to reuse _bind_ports and update the function for the case where number of sandboxes are bigger than number of ports to bind.

rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -55,24 +58,64 @@ def create_routers_bind_ports(self, router_create_args=None,
                                   port_create_args=None,
                                   ports_per_network=None,
                                   port_bind_args=None):
-
+        overlay_setup = router_create_args.get("overlay", True)
```
hzhou8: It seems unnecessary for a new argument for overlay_setup, since routers are already about overlay.

rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -67,6 +67,10 @@ def create_routers_bind_ports(self, router_create_args=None,
 
         # Create ports on the logical networks
         sandboxes = self.context["sandboxes"]
+        if not sandboxes:
+            # We need to get all the sandbox on all farms for random
```
hzhou8: The comment is misleading. Whether or not to do random distribution doesn't rely on this condition "if not sandboxes". The change here is about "when there is no sandbox specified, we bind ports on all sandboxes". (note that, even when using all sandboxes, it still can be even distribution if number of lports is bigger than number of sandboxes.

hzhou8: on L74, the warning "Number of ports less than chassis: inbalance binding\n" should be updated to:
"Number of ports less than chassis: random binding\n"

noah8713: Sure will correct the comments and message.

hzhou8:
LGTM, thanks!

PR #138, opened 2018-02-02
Use 'router' as address for the lsp connected to logical router.

hzhou8:
In history the mac of the lsp connected to logical router must be
the same as the mac of the peer logical router port, which is redundant.
Now that value 'router' is supported, we shall use it directly.


l8huang (approved):
LGTM

PR #143, opened 2018-02-06
prepare.sh: Ignore the error from grep if the user cannot be found

jsitnicki:
Otherwise the non-zero exit status from a subshell will cause the script to fail which is not what we want since we expect that the user might not exist yet.


hzhou8 (approved):
LGTM. Thanks Jakub!

PR #146, opened 2018-02-16
Re-use existing lswitches when new computes are added:

noah8713:
1. No need to create LRs and LS when computes get added; instead create lports in the same LRs and LS and bind them to the computes.
2. Make sure to pass the sandbox tag to load all new Sandboxes of that farm.
3. Add use_existing_routers flag to load existing northd data.

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients.py
+++ b/rally_ovs/plugins/ovs/ovsclients.py
@@ -101,6 +101,11 @@ def get_lswitch_info(info):
 
     return lswitches
 
+def parse_lswitch_list(lswitch_data):
+    lswitches = []
+    for line in lswitch_data.splitlines():
+        lswitches.append({"name": line.split(" ")[1].strip('()')})
+    return lswitches
 
 def set_colval_args(*col_values):
     args = []
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -12,7 +12,6 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-import pipes
 import sys
 import itertools
 import pipes
@@ -122,9 +121,11 @@ class OvnNbctl(OvsClient):
             self.run("ls-del", args=params)
 
 
-
         def lswitch_list(self):
-            self.run("ls-list")
+            stdout = StringIO()
+            self.run("ls-list", stdout=stdout)
+            output = stdout.getvalue()
+            return parse_lswitch_list(output)
 
         def lswitch_port_add(self, lswitch, name):
             params =[lswitch, name]
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -82,7 +82,7 @@ class OvnScenario(scenario.OvsScenario):
         ovn_nbctl = self.controller_client("ovn-nbctl")
         ovn_nbctl.set_sandbox("controller-sandbox")
         ovn_nbctl.enable_batch_mode(False)
-        ovn_nbctl.lswitch_list()
+        return ovn_nbctl.lswitch_list()
 
     @atomic.action_timer("ovn.delete_lswitch")
     def _delete_lswitch(self, lswitches):
--- a/rally_ovs/plugins/ovs/scenarios/ovn_network.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn_network.py
@@ -12,6 +12,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import netaddr
 from rally.common import logging
 from rally_ovs.plugins.ovs.scenarios import ovn
 from rally_ovs.plugins.ovs import utils
@@ -88,7 +89,20 @@ class OvnNetwork(ovn.OvnScenario):
 
         sandboxes = self.context["sandboxes"]
 
-        lswitches = self._create_networks(network_create_args)
+        use_existing_networks = port_create_args. \
+            get("use_existing_networks", False)
+        if not use_existing_networks:
+            lswitches = self._create_networks(network_create_args)
+        else:
+            lswitches = self._list_lswitches()
+            num_switches = len(lswitches)
+            start_cidr = network_create_args.get("start_cidr", "")
+            if start_cidr:
+                start_cidr = netaddr.IPNetwork(start_cidr)
+                for i in range(num_switches):
+                    if start_cidr:
+                        lswitches[i]["cidr"] = start_cidr.next(i)
+
         for lswitch in lswitches:
             lports = self._create_lports(lswitch, port_create_args, ports_per_network)
             self._bind_ports(lports, sandboxes, port_bind_args)


noah8713:
Fix submitted at upstream and merged @https://github.com/openstack/rally/commit/dbfae0ab4801d7eefdbec3d2df10457098532918

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -67,11 +67,11 @@ def run(self, cmd, opts=[], args=[], stdout=sys.stdout, stderr=sys.stderr):
                 cmd = itertools.chain(cmd_prefix, ["ovn-nbctl"], opts, [cmd], args)
                 self.cmds.append(" ".join(cmd))
 
-            self.ssh.run("\n".join(self.cmds),
-                         stdout=stdout, stderr=stderr)
+            exit_status, data = self.ssh.run(
```
hzhou8: this is not necessary, since stdout already contains the output. If we do this, we need to update the requirement.txt to ensure the latest version of rally is used so that "data" is returned from self.ssh.run().

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -122,9 +122,8 @@ def lswitch_del(self, name):
             self.run("ls-del", args=params)
 
 
-
         def lswitch_list(self):
-            self.run("ls-list")
+            return self.run("ls-list | awk '{print $2}'")
```
hzhou8: we should not change the behavior of lswitch_list. Parsing the information can be done in the caller, according to the caller's needs. For example, in this patch in ovn._list_lswitches() it anyway performances extra parsing, so this awk here is not very valuable while losing information of the original command output.

rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -56,14 +56,24 @@ def create_routers_bind_ports(self, router_create_args=None,
                                   ports_per_network=None,
                                   port_bind_args=None):
 
-        # Create routers and logical networks, and connect them
```
hzhou8: Instead of modifying create_routers_bind_ports scenario, it is better to keep it but modify the create_and_bind_ports scenario for this purpose, because the purpose is to create and bind ports in existing lswitches, and have nothing to do with lrouters.

The arg "use_existing_routers" can be changed to "use_existing_lswitches" as port_bind_args instead of router_create_args.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -351,8 +355,9 @@ def _bind_ports(self, lports, sandboxes, port_bind_args):
 
         LOG.info("Bind lports method: %s" % self.install_method)
         install_method = self.install_method
-
```
hzhou8: This change is not needed, since the logic of sandbox selection is not changed. If we want the behavior of doing random selection even when lports > sandboxes, we can add another arg "force_random_sandbox_selection" to control this behavior.

hzhou8 (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn.py
noah8713: sure!

rally_ovs/plugins/ovs/scenarios/ovn_network.py
noah8713: @hzhou8 I think it's fine because are just addressing overlay use case here.  We dont want to confuse bridge workloads here

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: It is debatable as this just gives the results what we need. Anyways we will have to parse lswitches data. So better do it at a calling time!

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: @hzhou8 So stdout was returning None. That is the reason this is needed

hzhou8: for stdout to return the output, you need to provider the parameter. e.g.:
my_stdout = StringIO()
self.run(stdout=my_stdout)

rally_ovs/plugins/ovs/scenarios/ovn_network.py
hzhou8: It is not about bridge/overlay. The function create_and_bind_port is generic for both use cases, it just doesn't create lrouters. So it's better to avoid tight couple it to specific assumptions.

rally_ovs/plugins/ovs/ovsclients_impl.py
hzhou8: > It is debatable as this just gives the results what we need.
The lswitch uuids could be useful for other callers, so I'd suggest to keep the original information. If in every caller we end up doing same processing to filter out the uuids and only need names, it makes sense to pre-precess then.

rally_ovs/plugins/ovs/ovsclients.py
```diff
@@ -101,6 +101,11 @@ def get_lswitch_info(info):
 
     return lswitches
 
+def get_lswitch_by_name(lswitch_data):
```
hzhou8: function named as ..._by_name usually means the input is a name, and it returns the results for the given name. This function is for parsing ls-list output, so please rename it to parse_lswitch_list.

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -41,6 +40,7 @@ def __init__(self, credential):
             self.sandbox = None
             self.batch_mode = False
             self.cmds = None
+            self.cred = credential
```
hzhou8: This is unrelated change, please remove

rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -88,7 +88,16 @@ def create_and_bind_ports(self,
 
         sandboxes = self.context["sandboxes"]
 
-        lswitches = self._create_networks(network_create_args)
+        # For scale test, we need to also add more ports on demand by adding
+        # new HVs. Hence, we can skip creating lswitches and lrouters since
```
hzhou8: the comment "lrouters" is misleading here. In fact the comment is not necessary, since the arg "use_existing_networks" is self-explaining.

rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -88,7 +88,16 @@ def create_and_bind_ports(self,
...
+        # we just need lports.
+        use_existing_networks = port_create_args. \
+            get("use_existing_networks", False)
+        if not use_existing_networks:
+            lswitches = self._create_networks(network_create_args)
+        else:
+            lswitches = self._list_lswitches()
```
hzhou8, line 97: _list_lswitches only return ls names, but here we need not only names but also CIDRs. Without CIDRs, lports won't have IP addresses and the ARP flows won't be created, and so the test result can be different from production environment because the number of flows will be much less. We need to solve this problem.

hzhou8 (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn_network.py
noah8713: sure

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: ok I will remove

rally_ovs/plugins/ovs/ovsclients.py
noah8713: ok will do

noah8713: done

rally_ovs/plugins/ovs/scenarios/ovn_network.py
noah8713: done

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: done

rally_ovs/plugins/ovs/scenarios/ovn_network.py
noah8713: Thats taken care!

hzhou8 (approved):

PR #147, opened 2018-02-19
Fix parser for ovn-nbctl show command output

jsitnicki:
Since openvswitch v2.6.0 output format of `ovn-nbctl show` has
changed. Any occurrences of "lswitch" have been replaced with "switch",
and occurrences of "lport" have replaced with just "port". Adapt to
the changes.

Ideally we shouldn't be doing output scraping if output format is
free-form but instead ask for machine readable output like JSON or
CSV.

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients.py
+++ b/rally_ovs/plugins/ovs/ovsclients.py
@@ -91,11 +91,11 @@ def get_lswitch_info(info):
     lswitch = None
     for line in info.splitlines():
         tokens = line.strip().split(" ")
-        if tokens[0] == "lswitch":
+        if tokens[0] == "switch":
             name = tokens[2][1:-1]
             lswitch = {"name":name, "uuid":tokens[1], "lports":[]}
             lswitches.append(lswitch)
-        elif tokens[0] == "lport":
+        elif tokens[0] == "port":
             name = tokens[1][1:-1]
             lswitch["lports"].append({"name":name})
 


jsitnicki:
@hzhou8 out of curiosity, do you need ovn-scale-test to work with openvswitch <= 2.6.0?

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -186,6 +186,8 @@ def acl_del(self, lswitch):
             params = [lswitch]
             self.run("acl-del", args=params)
 
+        # TODO: Don't try to process free-form output from `ovn-nbctl show`.
+        #       Ask for machine readable output, in JSON or CSV, instead.
```
hzhou8: Good point. Ideally we can use OVSDB python lib directly, but it might be a little bit overkill. For this ovsclients_impl I think the show command has nothing TODO here. I'd suggest to remove the TODO and create an issue for using OVSDB python lib.

hzhou8:
@jsitnicki it seems not much value to support <= 2.6 since OVN was immature back at that time.

rally_ovs/plugins/ovs/ovsclients_impl.py
jsitnicki: Agree, let me drop it and open an issue. For now using `ovn-nbctl` is sufficient for generating the workload but likely it will become a bottleneck as we increase the load. OpenStack Neutron plugin for OVN has started out using `ovn-nbctl` but then had to switch to ovsdbapp lib because the overhead of spawning a process and receiving whole database contents for each operation, like a adding a port, was too costly.

jsitnicki:
@hzhou8 I've addressed your comments, can you please give it another look?

jsitnicki:
Opened issue #153.

hzhou8:
LGTM. Thanks @jsitnicki !

PR #148, opened 2018-02-19
Make OvnNorthbound context class name unique

jsitnicki:
Fix a naming conflict with scenarios, where we also have a class named
OvnNorthbound. Before it was working by accident, due to a spelling
mistake in OvnNorthbound context class name.


hzhou8:
LGTM

PR #149, opened 2018-02-19
Correct timer name for list ports action in OvnScenario

jsitnicki:
Timer name doesn't correspond to action name. Fix it.


hzhou8 (approved):
LGTM

PR #150, opened 2018-02-19
Don't fail if the start CIDR was not given when creating an lswitch

jsitnicki:
"cidr" key will be missing if start_cidr was not provided. Handle it.


hzhou8 (approved):
LGTM

PR #154, opened 2018-03-04
Enable ssl for scale test which includes:

noah8713:
1. HVs ovn-controller talk to southbound db via ssl on port 6642
2. Starting and setting of southbound and northbound dbs with pssl on ports
   6642 and 6641 respectively.


rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
```diff
@@ -650,31 +655,67 @@ OVN_SB_DB=unix:$sandbox/db-sb.sock; export OVN_SB_DB
...
+                    --pidfile=$prog_name.pid \
+                    --unixctl=$prog_name.ctl \
+                    -vconsole:off -vsyslog:off -vfile:info \
+                    --log-file=$prog_name.log \
+                    --remote="p$OVN_SB_DB" \
+                    --remote=db:Open_vSwitch,Open_vSwitch,manager_options \
+                    conf-sb.db ovnsb.db
+                pid=`cat $sandbox_name/$prog_name.pid`
+                mv $sandbox_name/$prog_name.ctl $sandbox_name/$prog_name.$pid.ctl
+            fi
```
hzhou8, line 718: Could you abstract the different part only, instead of a big if-else block?

rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
```diff
@@ -694,19 +735,35 @@ EOF
...
     # Initialize database.
     if $controller ; then
-        init_ovsdb_server "ovsdb-server-nb" $OVN_NB_DB
-        init_ovsdb_server "ovsdb-server-sb" $OVN_SB_DB
+        if $enable_ssl ; then
+            tar -xzvf certs.tar.gz
+            abs_path=`pwd`/certs
+            ovn-nbctl set-ssl $abs_path/ovnnb-privkey.pem  $abs_path/ovnnb-cert.pem \
+                $abs_path/cacert.pem
+            ovn-nbctl set-connection pssl:6641:127.0.0.1
```
hzhou8, line 743: why 127.0.0.1? Rally may need to access NB remotely.

rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
```diff
@@ -694,19 +735,35 @@ EOF
 
     # Initialize database.
     if $controller ; then
-        init_ovsdb_server "ovsdb-server-nb" $OVN_NB_DB
-        init_ovsdb_server "ovsdb-server-sb" $OVN_SB_DB
```
hzhou8, line 698: Why init_ovsdb_server is not needed?

rally_ovs/plugins/ovs/scenarios/sandbox.py
```diff
@@ -159,10 +160,16 @@ def _create_sandbox(self, sandbox_create_args):
...
+                if ssl:
+                    cmd = "./ovs-sandbox.sh --ovn --controller-ip %s \
+                                 --host-ip %s/%d --device %s --ssl" % \
+                          (controller_ip, host_ip, sandbox_cidr.prefixlen,
+                           net_dev)
+                else:
+                    cmd = "./ovs-sandbox.sh --ovn --controller-ip %s \
+                                 --host-ip %s/%d --device %s" % \
+                             (controller_ip, host_ip, sandbox_cidr.prefixlen,
+                                    net_dev)
```
hzhou8, line 172: please abstract out the different part only (--ssl)

hzhou8 (changes requested):

rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
noah8713: So by default it will be localhost where we are setting up controller sandbox. We are not using cluster here. Hence when we run all the nbctl commands too , we source sandbox rc and do it. Once we introduce to spin up clustered db, we can leverage the actual IPs. Not needed now.

rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
noah8713: sure

rally_ovs/plugins/ovs/scenarios/sandbox.py
noah8713: sure

PR #160, opened 2018-04-05
Enable support to get openflow count from all HVs

noah8713:


rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -413,7 +413,17 @@ def _wait_up_port(self, lports, wait_sync, install_method):
...
+        sandbox_num = len(sandboxes)
+        install_method = self.install_method
+        for j in range(sandbox_num):
+            sandbox = sandboxes[j]["name"]
+            farm = sandboxes[j]["farm"]
+            ovs_ofctl = self.farm_clients(farm, "ovs-ofctl")
+            ovs_ofctl.set_sandbox(sandbox, install_method)
+            bridge = sandbox_args.get('bridge', 'br-int')
+            lflow_count = ovs_ofctl.dump_flows(bridge)
+            LOG.debug('openflow count on %s is %s' % (sandbox, lflow_count))
```
hzhou8: How does it work without return value? Just debug log?

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -289,6 +289,82 @@ def db_set(self, table, record, *col_values):
             self.run("set", args=args)
 
     def create_client(self):
-        print "*********   call OvnNbctl.create_client"
```
hzhou8: Unrelated change. Please submit a new patch.

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -289,6 +289,82 @@ def db_set(self, table, record, *col_values):
...
+    class _OvsOfctl(object):
+
+        def __init__(self, credential):
+            self.ssh = get_ssh_from_credential(credential)
+            self.context = {}
+            self.batch_mode = False
+            self.sandbox = None
+            self.cmds = None
+
+        def enable_batch_mode(self, value=True):
```
hzhou8: ofctl doesn't have batch mode. please remove all batch mode related code.

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -289,6 +289,82 @@ def db_set(self, table, record, *col_values):
...
+            self.batch_mode = bool(value)
+
+        def set_sandbox(self, sandbox, install_method="sandbox"):
+            self.sandbox = sandbox
+            self.install_method = install_method
+
+        def run(self, cmd, opts=[], args=[], stdout=sys.stdout, stderr=sys.stderr):
+            self.cmds = self.cmds or []
+
+            # TODO: tested with non batch_mode only for docker
```
hzhou8: This is misleading. Please don't add docker related code if not tested. A TODO comment can be added for supporting docker in the future.

hzhou8 (changes requested):

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: Ok will do

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: Sure!

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: ok I will address that in separate patch

rally_ovs/plugins/ovs/scenarios/ovn.py
noah8713: Yeah just debug logs for now. Can extend it to write to file?

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -413,7 +413,17 @@ def _wait_up_port(self, lports, wait_sync, install_method):
         if wait_sync != "none":
             ovn_nbctl.sync(wait_sync)
 
-
-
-
+    @atomic.action_timer("ovn_network.list_oflow_count_on_all_sandboxes")
+    def _list_oflow_count_on_all_sandboxes(self, sandboxes,
+                                           sandbox_args):
```
jsitnicki: Name is a bit confusing. We're actually listing flow count on given sandboxes, right? Not all sandboxes in the deployment.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -413,7 +413,17 @@ def _wait_up_port(self, lports, wait_sync, install_method):
...
+        sandbox_num = len(sandboxes)
+        install_method = self.install_method
+        for j in range(sandbox_num):
```
jsitnicki: Why not just iterate over values and avoid `sandbox_num` all together. I think that would be more idiomatic.

rally_ovs/plugins/ovs/scenarios/ovn.py
jsitnicki: Scenario can report arbitrary data, which becomes part of the report:

https://rally.readthedocs.io/en/latest/task/index.html#tab-scenario-data

I would suggest looking into using this API:

https://github.com/openstack/rally/blob/master/rally/task/scenario.py#L146

Logs are not really intended for this.

rally_ovs/plugins/ovs/scenarios/ovn.py
noah8713: @jsitnicki it's for each SBs.

rally_ovs/plugins/ovs/scenarios/ovn.py
noah8713: Sure that sounds fine too! :)

rally_ovs/plugins/ovs/scenarios/ovn.py
noah8713: @jsitnicki  Ack! Will give it a try

noah8713: Done!

rally_ovs/plugins/ovs/ovsclients_impl.py
noah8713: Sure! will submit in separate patch

noah8713:
Tested with creating 2 sandboxes and one controller which created 6 total openflows on each sandbox.

Works as expected below:
--------------------------------------------------------------------------------
Task 6eed038b-ac5f-49e8-913d-d741cd1e5e96: finished
--------------------------------------------------------------------------------

test scenario OvnNetwork.list_oflow_count_on_all_sandboxes
args position 0
args values:
{
  "runner": {
    "type": "serial",
    "times": 1
  },
  "context": {
    "sandbox": {},
    "ovn_multihost": {
      "controller": "ovn-new-controller-node"
    }
  }
}
+---------------------------------------------------------------------------------------------------------------------+
|                                                Response Times (sec)                                                 |
+------------------------------------------------+-------+--------+--------+--------+-------+-------+---------+-------+
| action                                         | min   | median | 90%ile | 95%ile | max   | avg   | success | count |
+------------------------------------------------+-------+--------+--------+--------+-------+-------+---------+-------+
| ovn_network.list_oflow_count_for_all_sandboxes | 0.354 | 0.354  | 0.354  | 0.354  | 0.354 | 0.354 | 100.0%  | 1     |
| total                                          | 0.361 | 0.361  | 0.361  | 0.361  | 0.361 | 0.361 | 100.0%  | 1     |
+------------------------------------------------+-------+--------+--------+--------+-------+-------+---------+-------+

Scenario Specific Results

+-------------------------------------------------------------------+
|           Openflow count on each sandbox in StackedArea           |
+----------------------+-----+--------+--------+--------+-----+-----+
| key                  | min | median | 90%ile | 95%ile | max | avg |
+----------------------+-----+--------+--------+--------+-----+-----+
| sandbox-192.230.64.0 | 6.0 | 6.0    | 6.0    | 6.0    | 6.0 | 6.0 |
| sandbox-192.230.64.1 | 6.0 | 6.0    | 6.0    | 6.0    | 6.0 | 6.0 |
+----------------------+-----+--------+--------+--------+-----+-----+

Load duration: 0.366648197174
Full duration: 0.391592979431

rally_ovs/plugins/ovs/scenarios/ovn.py
jsitnicki: OK, I think I follow now.

The scenario processes all sandboxes (`OvnNetwork.list_oflow_count_on_all_sandboxes`).

The atomic action processes the sandboxes passed in the arguments (`Ovn._list_oflow_count_for_all_sandboxes`)

jsitnicki:
Awesome, LGTM!

jsitnicki:
Actually, one more comment.

I would consider if this new scenario doesn't fit better into `OvnSandbox` module.

noah8713:
Thanks @jsitnicki for review.
You mean move the lflow logic to ovn_sandbox? The reason I choose current place is because getting openflow count is related to ovs scenarios, similar to other ovn/ovs client commands. Hence, I feel this would be ok. Let me know

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -292,3 +292,67 @@ def create_client(self):
...
+            self.cmds = None
+
+
+        def set_sandbox(self, sandbox, install_method="sandbox"):
+            self.sandbox = sandbox
+            self.install_method = install_method
+
+        def run(self, cmd, opts=[], args=[], stdout=sys.stdout, stderr=sys.stderr):
+            # TODO: add support for docker
+            self.cmds = self.cmds or []
```
hzhou8: no batch mode for ovs-ofctl, so self.cmds is useless for this class. The run() function should simply self.ssh.run().

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -292,3 +292,67 @@ def create_client(self):
...
+
+            cmd = itertools.chain(["ovs-ofctl"], opts, [cmd], args)
+            self.cmds.append(" ".join(cmd))
+
+            self.ssh.run("\n".join(self.cmds),
+                         stdout=stdout, stderr=stderr)
+
+            self.cmds = None
+
+        def flush(self):
```
hzhou8: this function is not needed because there is no batch mode

rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -114,4 +114,10 @@ def bind_ports(self):
     def bind_and_unbind_ports(self):
         pass
 
-
+    @scenario.configure(context={})
+    def list_oflow_count_for_all_sandboxes(self, sandbox_args={}):
```
hzhou8: Better to rename the function as list_oflow_count_for_sandboxes, since it doesn't have to be all. It depends on the sandbox_args. Same for L417.

hzhou8 (changes requested):

hzhou8 (approved):

PR #162, opened 2018-04-12
Allow task to just create lrouters and lswitches 

noah8713:
as we now have feature to re-use existing routers/networks

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn_network.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn_network.py
@@ -81,6 +81,26 @@ class OvnNetwork(ovn.OvnScenario):
 
     @validation.number("ports_per_network", minval=1, integer_only=True)
     @scenario.configure(context={})
+    def create_routers_and_networks(self, router_create_args=None,
+                                  router_connection_method=None,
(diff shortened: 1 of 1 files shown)


rally_ovs/plugins/ovs/scenarios/ovn_network.py
```diff
@@ -79,6 +79,26 @@ def create_routers_bind_ports(self, router_create_args=None,
             self._bind_ports(lports, sandboxes, port_bind_args)
 
 
+    @validation.number("ports_per_network", minval=1, integer_only=True)
+    @scenario.configure(context={})
+    def create_routers_and_networks(self, router_create_args=None,
```
jsitnicki, line 84: This scenario looks exactly like `OvnNetwork.create_routers()`. What am I missing?

noah8713: Correct! Thanks for the pointer. I was actually working on the code where I will distribute lports evenly on each HVs so that we can verify oflow count on all HV is same. Because in overlay it's tough to do so. This is not yet to be merge patch. More enhancement to come. So I will skip this part for now!

PR #165, opened 2018-04-25
Datapath context

jsitnicki:
These changes introduce a new context that can be used to pre-setup logical datapaths before the test run. It will create a given amount of logical routers, logical switches, and optionally connect the switches to the routers.

First we extract the code needed to create the routers, the switches, and connect them into a mixin, so that it can be shared between the scenarios and the contexts. Then the new context is added, with a minimal amount of code thanks to the extracted mixin.

We also add tests to cover the newly added context as well as the code moved around.

### DIFF
--- a/rally_ovs/plugins/ovs/context/datapath.py
+++ b/rally_ovs/plugins/ovs/context/datapath.py
@@ -1,0 +1,70 @@
+# Copyright 2018 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+# TODO(jkbs):
+# - Use constants for order
+# - Require ovn_multihost context when common validators are available
+
+from rally import consts
+from rally.common import logging
+from rally.task import context
+
+from rally_ovs.plugins.ovs import ovnclients
+
+
+LOG = logging.getLogger(__name__)
+
+
+@context.configure(name="datapath", order=115)
+class Datapath(ovnclients.OvnClientMixin, context.Context):
+    """Create datapath resources.
+
+    This context creates logical routers.
+    """
+
+    CONFIG_SCHEMA = {
+        "type": "object",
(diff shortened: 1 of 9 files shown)


hzhou8:
@jsitnicki Thanks for the PR. The Datapath context looks beautiful and also thanks for the great test coverage! I haven't gone through all the details yet, but I have a major comment here about the use of mixins. It seems the purpose of the mixin is to reuse the functions that were in scenarios only. I would suggest to abstract those functions to a class in a new module e.g. class OvnClient in ovn_client.py, and this client can be used by both scenarios and contexts as a property rather than mixin. Functionally, it might be the same, but I feel this make the structure more clear. What do you think?

jsitnicki:
@hzhou8 Thanks for the initial review!

Having a helper class, like `OvnClient`, would be simpler than having a mix-in, I agree. That was my initial approach. What made me abandon it was that at first I could not find a good way to share the code that instantiates `OvsClient`'s (that is now in `ClientsMixin`) without using a mix-in, because the existing scenarios rely on being able to call `controller_client` or `farm client`.

Then, when I moved on to making higher level routines shared (like `_create_lswitches`), I wanted to avoid having two instances of `Clients` cache in a scenario, since a standalone class like `OvnClient` would also need to instantiate `OvsClient`'s.

I can look into rewriting it into something where we have two `Clients` instances, along the lines of:
```python
class OvnClient(ClientsMixin):
    def __init__(self, context):
        # ...
    def create_lswitches(self, lswitch_create_args, num_switches=-1):
        #...

class OvnScenario(OvsScenario):
    def __init__(self, context):
        self.ovn_client = OvnClient(context)

class OvnNorthbound(OvnScenario):
    @scenario.configure(context={})
    def create_and_list_lswitches(self, lswitch_create_args=None):
        ovn_client.create_lswitches(lswitch_create_args)
        self._list_lswitches()
```
But I'm not sure if it's less or more confusing.

Anyway, does this reflect what you had in mind?

hzhou8:
@jsitnicki Thanks for the explain. I admit the instantiation part is trickier than I thought, so now I am fine with the mixin approach. Let me have a more detailed review for the rest of the patches.

@l8huang @noah8713 please take a look, too, in case I miss something.

hzhou8:
+1. It would be better if you could add/update samples for using datapath context.

jsitnicki:
@hzhou8 OK, let me come up with a couple new example scenario configs for the samples. One for pre-creating just switches, and another, more advanced, for pre-creating router(s) and switches. I'm thinking nothing complicated, a `Dummy.dummy` scenario with a `datapath` context. User can examine the logical datapath with `ovn-nbctl show` after the test run to see the effects.

noah8713:
Sorry was running late on reviews. :( Will take a look today!

l8huang:
LGTM

PR #167, opened 2018-05-02
Generalize _create_acl routine

LorenzoBianconi:
In order to add the capability to specify a non-constant ACL match value,
add 'match' key to acl_create_args dictionary. match key will be used to
specify the used address-set for a given ACL rule

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -186,7 +186,19 @@ class OvnScenario(scenario.OvsScenario):
         direction = acl_create_args.get("direction", "to-lport")
         priority = acl_create_args.get("priority", 1000)
         action = acl_create_args.get("action", "allow")
+        address_set = acl_create_args.get("address_set", "")
 
+        '''
+        match template: {
+            "direction" : "<inport/outport>",
+            "lport" : "<swicth port>",
+            "address_set" : "<address_set id>"
+            "l4_port" : "<l4 port number>",
+        }
+        '''
+        match_template = acl_create_args.get("match",
+                                             "%(direction)s == %(lport)s && \
+                                             ip4 && udp && udp.src == %(l4_port)s")
         if direction == "from-lport":
             p = "inport"
         else:
@@ -197,10 +209,11 @@ class OvnScenario(scenario.OvsScenario):
         ovn_nbctl.enable_batch_mode()
         for lport in lports:
             for i in range(acls_per_port):
-                match = '%s == "%s" && ip4 && udp && udp.src == %d' % \
-                        (p, lport["name"], 100 + i)
+                match = match_template % { 'direction' : p,
+                                           'lport' : lport["name"],
+                                           'address_set' : address_set,
+                                           'l4_port' : 100 + i }
                 ovn_nbctl.acl_add(sw, direction, priority, match, action)
-
             ovn_nbctl.flush()
 
 


rally_ovs/plugins/ovs/scenarios/ovn_nb.py
```diff
@@ -110,6 +110,7 @@ def create_and_list_acls(self,
         lswitches = self.get_or_create_lswitch_and_lport(lswitch_create_args,
                                     lport_create_args, lports_per_lswitch)
 
+        acl_create_args["match": "ip4 && udp"]
```
hzhou8: Sorry that I don't understand the syntax here. Is it supposed to be acl_create_args["match"] = "ip4 && udp"? Please forgive me if it turns out to be just my pool python skill :) At least it doesn't work in my local testing: 

>>> a = {'a': 'b'}
>>> a['a': 'c']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type

If the purpose is the same, then another question is: why overriding without checking the user configured args? We wouldn't need the args if we just hard code the match this way, right?

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -196,11 +197,11 @@ def _create_acl(self, lswitch, lports, acl_create_args, acls_per_port):
         ovn_nbctl.set_sandbox("controller-sandbox", self.install_method)
         ovn_nbctl.enable_batch_mode()
         for lport in lports:
+            match = '%s == "%s"' % (p, lport["name"])
+            if extra_match_val:
+                match += " && %s" % extra_match_val
             for i in range(acls_per_port):
-                match = '%s == "%s" && ip4 && udp && udp.src == %d' % \
-                        (p, lport["name"], 100 + i)
                 ovn_nbctl.acl_add(sw, direction, priority, match, action)
```
hzhou8: 1) It is better avoid adding identical ACLs, e.g. use the old way to change source port.

2) what if extra_match_val is empty? It would better to generate some default match conditions still like the old way

hzhou8 (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn.py
jsitnicki: What if we turn the `acl_create_args["match"]` into a meta-match that will be filled with values as needed, e.g.
```
In [9]: "%(direction)s == %(lport)s && udp.src == %(l4_port)s" % {'direction': 'inport', 'lport': 'foo', 'l4_port': 123}
Out[9]: 'inport == foo && udp.src == 123'

In [10]: "%(direction)s == %(lport)s" % {'direction': 'inport', 'lport': 'foo', 'l4_port': 123}
Out[10]: 'inport == foo'
```
And have the existing `match` as default value (`"%(direction)s == %(lport)s && ip4 && udp && udp.src == %(l4_port)d"`).

I think that should satisfy all needs.

rally_ovs/plugins/ovs/scenarios/ovn_nb.py
jsitnicki: Would be nice to add test coverage to scenarios that are being modified. It's not too much work with mocks. I can help if needed.

LorenzoBianconi: Yes sorry, just a typo

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -197,10 +209,11 @@ def _create_acl(self, lswitch, lports, acl_create_args, acls_per_port):
         ovn_nbctl.enable_batch_mode()
         for lport in lports:
             for i in range(acls_per_port):
-                match = '%s == "%s" && ip4 && udp && udp.src == %d' % \
-                        (p, lport["name"], 100 + i)
+                match = match_template % { 'direction' : p,
+                                           'lport' : lport["name"],
+                                           'address_set' : address_set,
```
hzhou8, line 214: Firstly, address_set is not in the match_template.
Secondly, if user didn't supply address_set (i.e. address_set == ""), we should not put it in the generated ACL

hzhou8 (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: Do you mean in the default match_template? The default template is the one used by create_and_list_acls() or create_and_delete_acls().
Since address_set is not in the default template, it is up to the user to provide both a match_template containing an address_set reference and a valid address_set id. If the user provides a templates not containing an address_set reference, it will not be put in the ACL rule. Am I missing something?

hzhou8: Now I got it. Thanks for explain. LGTM!

hzhou8 (approved):

PR #168, opened 2018-05-09
Redirect output of DdCtlMixin.get routine

LorenzoBianconi:
Redirect output of get DdCtlMixin.get routine from stdoutput to a
local variable

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients.py
+++ b/rally_ovs/plugins/ovs/ovsclients.py
@@ -18,6 +18,7 @@ import six
 
 from rally.common.plugin import plugin
 from utils import py_to_val
+from io import StringIO
 
 _NAMESPACE = "ovs"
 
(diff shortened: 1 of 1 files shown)


jsitnicki:
I'm wondering if in the end this is really needed. I've taken a look at the use of this change in https://github.com/LorenzoBianconi/ovn-scale-test/commits/switch_per_node_workload and I see that you _recreate_ the address set to add addresses to it. We don't need to do that, and I expect this is not what a sane CMS will do. We can append addresses to a set in a simpler way:
```
ovn-nbctl add Address_Set addrset0 addresses 1.1.1.1
```
And conversly remove them easily:
```
ovn-nbctl remove Address_Set addrset0 addresses 1.1.1.1
```

LorenzoBianconi:
@jsitnicki: ok, fine. I can add 'add' routine to DdCtlMixin class.
@hzhou8 do you think it matters to redirect the output of 'get' routine to a variable or it is fine to send it to stdout? (If so I can close the PR)

hzhou8 (approved):
Yes, it makes sense to return the output for get() interface. Thanks for the fix.

PR #169, opened 2018-05-09
Fix set_colval_args parsing

LorenzoBianconi:
Fix set_colval_args parsing if col_values size is one, e.g:

ovn-nbctl get Address_Set <address_set_name> addresses

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients.py
+++ b/rally_ovs/plugins/ovs/ovsclients.py
@@ -110,7 +110,9 @@ def parse_lswitch_list(lswitch_data):
 def set_colval_args(*col_values):
     args = []
     for entry in col_values:
-        if len(entry) == 2:
+        if len(entry) == 1:
+            col, op, val = '', '', entry[0]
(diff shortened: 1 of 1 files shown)


jsitnicki:
This change also won't be needed if we would be using `ovn-nbctl add/remove`, as I mentioned in comment to https://github.com/openvswitch/ovn-scale-test/pull/168.

Besides, I think the problem here is that `ovn-nbctl get` expects a list of column names, not a list of (column, relational operator, value) tuples. That is - inappropriate helper used in `DdCtlMixing.get()` IMO.

hzhou8:
I agree with @jsitnicki that the current implementation of DdCtlMixin.get() is wrong, since the set_colval_args() function adds operator, which should not be there for get. (see man ovn-nbctl, or ovn-sbctl or ovs-vsctl). set_colval_args() is valid for wait_until() or set/create (which is implemented in PR #166). @LorenzoBianconi would you mind fix this?

PR #170, opened 2018-05-09
Fix IP configuration in _create_lports routine

LorenzoBianconi:
Fix IP parsing/configuration in _create_lports routine since it
currently set the network address instead of a valid IP. E.g:
if the cidr is 192.168.0.1/24 and lport_amount is one, the configured
IP address will be 192.168.0.0
Moreover add the possibility to configure different IP addresses
belonging to the same subnet if _create_lports routine is run multiple
times in a row (e.g. using datapath context)

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -60,7 +60,8 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
         pass
 
     @atomic.action_timer("ovn.create_lport")
-    def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1):
+    def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1,
+                       lport_ip_shift = 1):
         LOG.info("create %d lports on lswitch %s" % \
                             (lport_amount, lswitch["name"]))
 
@@ -74,13 +75,14 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
         network_cidr = lswitch.get("cidr", None)
         ip_addrs = None
         if network_cidr:
-            end_ip = network_cidr.ip + lport_amount
+            end_ip = network_cidr.ip + lport_amount + lport_ip_shift
             if not end_ip in network_cidr:
                 message = _("Network %s's size is not big enough for %d lports.")
                 raise exceptions.InvalidConfigException(
                             message  % (network_cidr, lport_amount))
 
-            ip_addrs = netaddr.iter_iprange(network_cidr.ip, network_cidr.last)
+            ip_addrs = netaddr.iter_iprange(network_cidr.ip + lport_ip_shift,
+                                            network_cidr.last)
 
         ovn_nbctl = self.controller_client("ovn-nbctl")
         ovn_nbctl.set_sandbox("controller-sandbox", install_method)


jsitnicki:
I think you are looking for `IPnetwork.iter_hosts()`:
```python
In [25]: list(netaddr.IPNetwork('10.0.0.0/29').iter_hosts())
Out[25]:
[IPAddress('10.0.0.1'),
 IPAddress('10.0.0.2'),
 IPAddress('10.0.0.3'),
 IPAddress('10.0.0.4'),
 IPAddress('10.0.0.5'),
 IPAddress('10.0.0.6')]
```
That seems better that applying an offset manually.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -114,13 +114,14 @@ def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1):
...
         if network_cidr:
-            end_ip = network_cidr.ip + lport_amount
+            end_ip = network_cidr.ip + lport_amount + lport_ip_shift
             if not end_ip in network_cidr:
                 message = _("Network %s's size is not big enough for %d lports.")
                 raise exceptions.InvalidConfigException(
                             message  % (network_cidr, lport_amount))
 
-            ip_addrs = netaddr.iter_iprange(network_cidr.ip, network_cidr.last)
+            ip_addrs = list(netaddr.iter_iprange(network_cidr.ip + lport_ip_shift,
```
jsitnicki: (Copying my comment an out-of-PR review here.)

This is a bad idea, see how long `list(netaddr.IPNetwork("10.0.0.0/8")` takes. We need to use iterators where possible.

LorenzoBianconi: Ack, thx for the suggestion

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -114,13 +114,13 @@ def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1):
...
         if network_cidr:
-            end_ip = network_cidr.ip + lport_amount
+            end_ip = network_cidr.ip + lport_amount + lport_ip_shift
             if not end_ip in network_cidr:
                 message = _("Network %s's size is not big enough for %d lports.")
                 raise exceptions.InvalidConfigException(
                             message  % (network_cidr, lport_amount))
 
-            ip_addrs = netaddr.iter_iprange(network_cidr.ip, network_cidr.last)
+            ip_addrs = netaddr.IPNetwork(network_cidr.ip + lport_ip_shift).iter_hosts()
```
hzhou8: iter_hosts() iterates the all IPs for the subnet CIDR, so adding a shift doesn't make sense here. For example:

>>> list(netaddr.IPNetwork("10.0.1.3/29").iter_hosts())
[IPAddress('10.0.1.1'), IPAddress('10.0.1.2'), IPAddress('10.0.1.3'), IPAddress('10.0.1.4'), IPAddress('10.0.1.5'), IPAddress('10.0.1.6')]

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -114,13 +114,13 @@ def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1):
         network_cidr = lswitch.get("cidr", None)
         ip_addrs = None
         if network_cidr:
-            end_ip = network_cidr.ip + lport_amount
+            end_ip = network_cidr.ip + lport_amount + lport_ip_shift
             if not end_ip in network_cidr:
```
hzhou8, line 79: If we want to skip .0, we should use network_cidr.first here instead of network_cidr.ip. Then the lport_ip_shift is needed only for skipping the first <lport_ip_shift> IPs

hzhou8 (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: Ack, reviewing the code I guess we could use iter_iprange() instead of iter_hosts() since we need to define a given IP range

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: Please correct me if I am wrong but I get the same result with network_cidr.first and network_cidr.ip (the first IP of the class, that is .0). Moreover I would add the possibility to skip a given set of IPs in the iprange since I would like to reuse _create_lports routine with datapath context from @jsitnicki and specify different IP ranges in multiple runs

hzhou8: Sorry, please ignore this comment.

LorenzoBianconi: @hzhou8 no worries. So is it fine to update the PR using iter_iprange and lport_ip_shift to specify different IP ranges during multiple runs?

hzhou8: Sounds good to me. iter_iprange() here is used as an iterator (L139), so I don't see performance issues.

LorenzoBianconi: @hzhou8 ok, fine. What about lport_ip_shift? does it work for you?

hzhou8: Yes, I think it should work. We can make it start from 1 by default, and provides the flexibility to specify other values for multiple runs.

hzhou8 (approved):

PR #173, opened 2018-05-22
Split timed operations up.

putnopvut:
The atomic action timer "ovn_network.bind_port" is used for the
_bind_port() member of OvnScenario. If the "wait_up" parameter is true,
then it calls into the wait_up_port() member, which has its own atomic
action timer "ovn_network.wait_port_up". The result of this is that
ovn_network.wait_port_up contributes towards the ovn_network.bind_port
timer.

This commit makes it so that the ovn_network.wait_port_up timer is not
part of the ovn_network.bind_port timer. This way, rally reports make it
more clear where time is being spent.

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -247,10 +247,7 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
 
         return lswitches
 
-
-
-    @atomic.action_timer("ovn_network.bind_port")
-    def _bind_ports(self, lports, sandboxes, port_bind_args):
+    def _bind_ports_and_wait(self, lports, sandboxes, port_bind_args):
         port_bind_args = port_bind_args or {}
         wait_up = port_bind_args.get("wait_up", False)
         # "wait_sync" takes effect only if wait_up is True.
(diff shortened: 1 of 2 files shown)


rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -261,13 +258,19 @@ def _bind_ports(self, lports, sandboxes, port_bind_args):
                 "Unknown value for wait_sync: %s. "
                 "Only 'hv', 'sb' and 'none' are allowed.") % wait_sync)
 
+        LOG.info("Bind lports method: %s" % self.install_method)
+        install_method = self.install_method
```
jsitnicki: Pulling up `install_method` works but there's also no reason why `_wait_port_up` can't access `self.install_method` directly. In fact, it seems to be the most common pattern to see throughout the codebase:
```
ovn_nbctl.set_sandbox("controller-sandbox", self.install_method)
```

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -247,9 +247,6 @@ def _create_networks(self, network_create_args, num_networks=-1):
 
         return lswitches
 
-
-
-    @atomic.action_timer("ovn_network.bind_port")
     def _bind_ports(self, lports, sandboxes, port_bind_args):
```
jsitnicki: I would probably consider a rename to reflect its true function to something like `bind_ports_and_wait`. We have only 3 callers so it won't be massive. Also, I understand that double-underscore prefix (`__method`) is a convention in Python for naming class-private members, so I would avoid that. Seems confusing.

putnopvut:
@jsitnicki new commit pushed up. It should address your findings.

jsitnicki (approved):
LGTM

hzhou8:
LGTM. Thanks!

PR #174, opened 2018-05-31
Adapt shell scripts for running over SSH without a pseduo-tty

jsitnicki:
Because we're not allocating a pseduo-tty when running scripts remotely,
the `PATH` is set only to `/usr/local/bin:/usr/bin`, as if we we ran:
```
$ ssh -T rally@<node> /home/rally/ovs-install.sh
```
Due to that the scripts cannot find the installed OVS/OVN daemons or
tools like `ip`.

### DIFF
--- a/rally_ovs/plugins/ovs/deployment/engines/ovs/install.sh
+++ b/rally_ovs/plugins/ovs/deployment/engines/ovs/install.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Adjust PATH for non-interactive SSH sessions.
+export PATH=$PATH:/usr/local/sbin:/usr/sbin
+
 OVS_REPO=$1
 OVS_BRANCH=$2
 OVS_USER=$3
--- a/rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
+++ b/rally_ovs/plugins/ovs/deployment/engines/ovs/ovs-sandbox.sh
@@ -17,6 +17,9 @@
 #set -e # exit on first error
 #set -x
 
+# Adjust PATH for non-interactive SSH sessions.
+export PATH=$PATH:/usr/local/sbin:/usr/sbin
+
 run() {
     (cd "$sandbox" && "$@") || exit 1
 }
--- a/rally_ovs/plugins/ovs/deployment/engines/ovs/prepare.sh
+++ b/rally_ovs/plugins/ovs/deployment/engines/ovs/prepare.sh
@@ -2,6 +2,9 @@
 
 set -e # exit on first error
 
+# Adjust PATH for non-interactive SSH sessions.
+export PATH=$PATH:/usr/local/sbin:/usr/sbin
+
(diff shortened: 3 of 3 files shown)


rally_ovs/plugins/ovs/deployment/engines/ovs/install.sh
```diff
@@ -71,6 +71,15 @@ function install_ovs {
     sudo chown $OVS_USER /usr/local/var/log/openvswitch
 }
 
+function path_prepend {
```
l8huang: I guess you can simply use 
`export PATH="/usr/sbin:/usr/local/sbin:$PATH"`
here, duplicate path is not harm ;)

`export` is for if other script is called in turn, the PATH still contains "/usr/sbin:/usr/local/sbin".

jsitnicki: Perhaps I did over-engineer it :-) I sure can switch to just setting the `PATH`. If we do that, I think we need to append the might-be-missing entries so as not to override system's default configuration, in case it matters.

That said, I'm still not convinced that adjusting the `PATH` in our scripts is the right solution. It feels like it should be done outside of the scripts but I wasn't able to come up with a way to do it.

hzhou8: I understand @l8huang 's suggestion about exporting $PATH with extra paths appending the existing ones. However, I didn't understand @jsitnicki 's reply. @jsitnicki could explain more about the "might-be-missing entries", maybe an example would help.

jsitnicki: Sure, sorry if I wasn't clear. By might-be-missing-entries I meant the ones we will be appending to `PATH` because we expect they might not be there yet. This is `/usr/local/sbin` and `/usr/sbin` in this case.

Why I think we need to append instead of prepend? User or sys admin can set `PATH` in `~/.bashrc` or `/etc/bashrc`, for example to point it to a custom location for OVS/OVN binaries. If we prepend to `PATH` we will enforce favouring the binaries that live in `*/sbin` where binaries from system packages or locally installed ones live.

jsitnicki: I haven't found any viable alternative to setting the `PATH` in the scripts.

The alternative options that I've considered are:
1. Setting PATH in `~/.ssh/environment` for `rally` user, requires `sshd` to have `PermitUserEnvironment` enabled which is not the default.
2. Setting PATH in `~/.bashrc` or sourcing `/etc/profile` from `~/.bashrc`.

Neither sounds simpler or cleaner to me than setting the `PATH` in the script itself so I will just do that.

jsitnicki:
Changes in v2:

* Switch to simply appending to `PATH` as suggested by @l8huang 
* Update commit description - issue is actually not related to pseudo-tty allocation.

PTAL

hzhou8 (approved):

PR #177, opened 2018-06-15
Use optional_action_timer instead of action_timer in _create_acl routine

LorenzoBianconi:
Substitute action_timer with optional_action_timer in _create_acl
as timer routine since in some cases (e.g OpenShift scenario) where
multiple ACLs are added for a given logical port could be useful
to estimate the time needed to configure the complete set of rules
instead of considering each ACL separately. If _create_acl is run
with standard parameters the timer is executed by default


LorenzoBianconi:
the code has been merged, so close the PR

PR #182, opened 2019-07-26
Fix counting of IGMP flows

dceara:


dceara:
Sorry, for the follow up pull request. I had forgotten that before I was using a custom OVN build. This should fix the IGMP flow counting and work with OVN upstream.

PR #188, opened 2020-02-09
OpenShift workload

LorenzoBianconi:
Introduce osh workload in order to simulate a typical OpenShift deployment:
- 1 logical router, 'n' logical switches each with 'm' logical switch ports
- address_set for each couple of logical switch port
- 2 acls for each logical switch port

### DIFF
--- a/rally_ovs/plugins/ovs/context/datapath.py
+++ b/rally_ovs/plugins/ovs/context/datapath.py
@@ -89,6 +89,7 @@ class Datapath(ovnclients.OvnClientMixin, context.Context):
             "routers": routers,
             "lswitches": lswitches,
         }
+        self.context["ovs-internal-ports"] = {}
 
     def cleanup(self):
         pass  # Not implemented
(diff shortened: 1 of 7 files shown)


hzhou8 (commented):
This patch specifies 2 ports as a group. Does it make sense to make it flexible to have N ports as a group instead of hard-coded by 2? Presumably, 2 ports as a group will not have good scale test result, while increasing the group size would scale better. So I think it is better to have this parameter configurable so that different group size can be tested and compared.

In addition, it should be more efficient to use port-group instead of address-set with ACL. Using port-group can reduce number of ACLs largely, when group size is big. Of course, this can be added as a separate test scenario.

hzhou8:
Let me now if you'd like to make the group size configurable. Otherwise the whole PR LGTM.

LorenzoBianconi:
> Let me now if you'd like to make the group size configurable. Otherwise the whole PR LGTM.

I addressed your first comment regarding # of lsp for each address_set. I will add port-group support on top of this PR.

Thanks

huikang:
Could you squash the commits? Thanks.

LorenzoBianconi:
> Could you squash the commits? Thanks.

Do you mean to squash all 10 the commits in a single one?

huikang:
> Do you mean to squash all 10 the commits in a single one?

Yes.

LorenzoBianconi:
> > Do you mean to squash all 10 the commits in a single one?
> 
> Yes.

done

hzhou8:
I am ok with separate commits for each logically individual patch. It is easier to review and also easier to revert in case there are problems of a specific commit. @huikang could you explain the benefit of squashing the commits?

PR #189, opened 2020-02-19
Add Openshift-like 'switch per node' bring-up scenario.

dceara:
Add new OVN scenario OvnNorthboundFakeMultinode which implements a test
case that brings up and configures ovn-fake-multinode nodes
(logical_switch, connect to logical_router, configure mgmt port, ping
from mgmt port to default gateway).

Also:
- Make sure all ovn-nbctl and ovs-ssh and ovs-vsctl connections are
  properly closed.
- Don't enable mcast snooping by default.
- Fix router port IP configuration and store it in the logical switch
  port info.
- Remove ovsclient.create_client logs that flood the output and bring no
  real benefit.
- Move create_routed_network implementation to ovn.py to avoid plugin
  load errors in rally.
- Enhance ovn_sandbox create_sandbox scenario to allow creating
  sandboxes based on iteration number.
- Add wait_sync=ping option for binding logical ports: will wait until
  the gateway is reachable.

Co-authored-by: Mark Michelson <mmichels@redhat.com>
Signed-off-by: Mark Michelson <mmichels@redhat.com>
Co-authored-by: Lorenzo Bianconi <lorenzo.bianconi@redhat.com>
Signed-off-by: Lorenzo Bianconi <lorenzo.bianconi@redhat.com>

### DIFF
--- a/rally_ovs/plugins/ovs/ovnclients.py
+++ b/rally_ovs/plugins/ovs/ovnclients.py
@@ -44,6 +44,10 @@ class OvnClientMixin(ovsclients.ClientsMixin, RandomNameGeneratorMixin):
         self._stop_daemon()
         return self._start_daemon()
 
+    def _get_gw_ip(self, network_cidr):
+        # Use the last IP in the CIDR as gateway IP.
+        return netaddr.IPAddress(network_cidr.last - 1)
+
     def _create_lswitches(self, lswitch_create_args, num_switches=-1):
         self.RESOURCE_NAME_FORMAT = "lswitch_XXXXXX_XXXXXX"
 
@@ -137,8 +141,10 @@ class OvnClientMixin(ovsclients.ClientsMixin, RandomNameGeneratorMixin):
         base_mac[3:] = ['00']*3
         mac = utils.get_random_mac(base_mac)
 
+        gw = self._get_gw_ip(network["cidr"])
+        lrouter_port_ip = '{}/{}'.format(gw, network["cidr"].prefixlen)
         lrouter_port = ovn_nbctl.lrouter_port_add(router["name"], network["name"], mac,
-                                                  str(network["cidr"]))
+                                                  lrouter_port_ip)
         ovn_nbctl.flush()
 
 
--- a/rally_ovs/plugins/ovs/ovnclients.py
+++ b/rally_ovs/plugins/ovs/ovnclients.py
@@ -55,7 +55,7 @@ class OvnClientMixin(ovsclients.ClientsMixin, RandomNameGeneratorMixin):
         if start_cidr:
             start_cidr = netaddr.IPNetwork(start_cidr)
 
-        mcast_snoop = lswitch_create_args.get("mcast_snoop", "true")
+        mcast_snoop = lswitch_create_args.get("mcast_snoop", "false")
         mcast_idle = lswitch_create_args.get("mcast_idle_timeout", 300)
         mcast_table_size = lswitch_create_args.get("mcast_table_size", 2048)
 
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -371,6 +371,18 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
         if wait_up:
             self._wait_up_port(lports, wait_sync)
 
+    @atomic.action_timer("ovn.bind_ovs_vm")
+    def _bind_ovs_port(self, lport, ovs_vsctl, internal=True):
+        port_name = lport["name"]
+        LOG.info("bind %s to %s on %s" % (port_name, sb_name, farm))
+
+        ovs_vsctl.add_port('br-int', port_name, internal=internal)
+        ovs_vsctl.db_set('Interface', port_name,
+                        ('external_ids', {"iface-id":port_name,
+                                            "iface-status":"active"}),
+                        ('admin_state', 'up'))
+
+    @atomic.action_timer("ovn.bind_internal_vm")
     def _bind_ovs_internal_vm(self, lport, sandbox, ovs_ssh):
         port_name = lport["name"]
         port_mac = lport["mac"]
@@ -465,16 +477,7 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
                 ovs_vsctl.set_sandbox(sandbox, self.install_method,
                                       sandbox_data['host_container'])
                 ovs_vsctl.enable_batch_mode()
-                port_name = lport["name"]
-                port_mac = lport["mac"]
-                port_ip = lport["ip"]
-                LOG.info("bind %s to %s on %s" % (port_name, sandbox, farm))
-
-                ovs_vsctl.add_port('br-int', port_name, internal=internal)
-                ovs_vsctl.db_set('Interface', port_name,
-                                 ('external_ids', {"iface-id": port_name,
-                                                   "iface-status": "active"}),
-                                 ('admin_state', 'up'))
+                self._bind_ovs_port(lport, ovs_vsctl, internal)
                 ovs_vsctl.flush()
 
                 # If it's an internal port create a "fake vm"
@@ -495,15 +498,7 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
                                       sandboxes[j]["host_container"])
                 ovs_vsctl.enable_batch_mode()
                 for index, lport in enumerate(lport_slice):
-                    port_name = lport["name"]
-
-                    LOG.info("bind %s to %s on %s" % (port_name, sandbox, farm))
-
-                    ovs_vsctl.add_port('br-int', port_name, internal=internal)
-                    ovs_vsctl.db_set('Interface', port_name,
-                                     ('external_ids', {"iface-id":port_name,
-                                                       "iface-status":"active"}),
-                                     ('admin_state', 'up'))
+                    self._bind_ovs_port(lport, ovs_vsctl, internal)
                     if index % 400 == 0:
                         ovs_vsctl.flush()
                 ovs_vsctl.flush()
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -98,8 +98,15 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
         pass
 
     @atomic.action_timer("ovn.create_lport")
-    def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1,
-                       lport_ip_shift = 1):
+    def _create_lports(self, lswitches, lport_create_args = [], lport_amount=1,
+                       lport_ip_shift=1):
+        lports = []
+        for lswitch in lswitches:
+            lports += self._create_switch_lports(lswitch, lport_create_args)
+        return lports
+
+    def _create_switch_lports(self, lswitch, lport_create_args = [],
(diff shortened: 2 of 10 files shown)


dceara:
CC:
@LorenzoBianconi 
@putnopvut

dceara:
CC: @hzhou8, it would be great if you could have a look. Thanks!

rally_ovs/plugins/ovs/context/ovn_nb.py
```diff
@@ -46,6 +46,7 @@ def setup(self):
 
         self.context["ovn-nb"] = lswitches
         self.context["ovs-internal-ports"] = {}
+        ovn_nbctl.close()
```
l8huang: The SSH connections to a controller/ farm node are cached(see [OvsClient.__call__()](https://github.com/ovn-org/ovn-scale-test/blob/3fbe332f96fe5881a05e0579e19e8c59f0d10b5a/rally_ovs/plugins/ovs/ovsclients.py#L53)) for reusing, so it shouldn't be closed, and next time calling self.controller_client() will return it. The purpose is improving the performance of running commands on a node by saving time to establish ssh connection. 

Once running a case is over, ssh connections will be closed before the process exit.

hzhou8:
> CC: @hzhou8, it would be great if you could have a look. Thanks!

Sorry for the slow reviewing. I was confused by the ssh closing. Please check @l8huang 's comment. Were you observing any ssh connection leak? How did it happen?

In addition, in the future (and in the past as well) I'd prefer one commit for one change instead of squashing many changes together. I know some repo's prefer one commit for each PR, but there can be multiple PRs in that case. From my perspective multiple commits in one PR is ok, but I am open to this. In either case, each change should be in a separated commit.

dceara:
@l8huang @hzhou8, thanks for reviewing this!

> > CC: @hzhou8, it would be great if you could have a look. Thanks!
> 
> Sorry for the slow reviewing. I was confused by the ssh closing. Please check @l8huang 's comment. Were you observing any ssh connection leak? How did it happen?
> 

We've been using ovn-scale-test lately to test large scale deployments and with one open connection per farm we ended up with lots of ssh sessions that live until the process exits. At some point we couldn't create more ssh sessions. However, we are testing with ovn-fake-multinode as "physical" deployment which essentially runs virtual nodes inside physical machines. I will look into reusing SSH sessions to only have one per physical machine.

> In addition, in the future (and in the past as well) I'd prefer one commit for one change instead of squashing many changes together. I know some repo's prefer one commit for each PR, but there can be multiple PRs in that case. From my perspective multiple commits in one PR is ok, but I am open to this. In either case, each change should be in a separated commit.

Sure, will do.

Thanks,
Dumitru

l8huang:
> At some point we couldn't create more ssh sessions.

what's the root cause of ssh session creation failure? Is it running out memory or file descriptor?

dceara:
> > At some point we couldn't create more ssh sessions.
> 
> what's the root cause of ssh session creation failure? Is it running out memory or file descriptor?

Running out of file descriptors.

```
Error SSHError: Exception <class 'paramiko.ssh_exception.SSHException'> was raised during connect to <host>:22. Exception value is: SSHException('No existing session')
```

We could increase the file descriptor limit but then we also hit an issue in rally itself because it uses `select` internally and doesn't work with more than 1K FDs.

I think, however, I can address all this by reusing the SSH connection for all containers that are spawned on a physical machine. I'll try it out and update the PR when I have it working.

Thanks,
Dumitru

dceara:
@hzhou8 @l8huang As suggested, I split the PR in multiple commits (one per logical change). For now I left out the SSH connection changes. 

I'll send a follow up PR once I have the code ready to reuse SSH connections for all containers spawned on the same physical machine. Until then we'll work around it on our setups by increasing the file descriptor limit when needed.

Thanks,
Dumitru

dceara:
@hzhou8 @l8huang Sorry, I had to force push the PR because I realized that in commit `Add wait_sync=ping option for binding logical ports.` I had forgotten to add an `import time`. This is the only change I made in the new version.

rally_ovs/plugins/ovs/ovnclients.py
```diff
@@ -44,6 +44,10 @@ def _restart_daemon(self):
         self._stop_daemon()
         return self._start_daemon()
 
+    def _get_gw_ip(self, network_cidr):
```
hzhou8, line 47: Before the change it was using the first IP to create LRP, now it uses the last IP, so what's the purpose of this change?

rally_ovs/plugins/ovs/scenarios/ovn_sandbox.py
```diff
@@ -96,7 +100,21 @@ def create_sandbox(self, sandbox_create_args=None):
...
+        farm = sandbox_create_args.get("farm", "")
+        if not farm:
+            iteration = self.context["iteration"]
+            farm_prefix = sandbox_create_args.get("farm-prefix", "")
+            farm = "{}{}".format(farm_prefix, iteration)
+            
+            # Increment sandbox cidr based on iteration.
+            sandbox_create_args = copy.copy(sandbox_create_args)
+            sandbox_create_args["start_cidr"] = "{}/{}".format(
+                str(start_cidr.ip + iteration + 1),
```
hzhou8: Should it be: start_cidr.ip +  iteration * <number of IPs in each iteration> + 1 ? Or each iteration only needs one IP? Could explain more about the use case?

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -487,20 +489,59 @@ def _bind_ports(self, lports, sandboxes, port_bind_args):
...
+        ovs_ssh.set_sandbox(sandbox, self.install_method,
+                            sandbox['host_container'])
+        ovs_ssh.enable_batch_mode(False)
+
+        success = False
+        for i in range(wait_timeout_s):
+            try:
+                ovs_ssh.run("ip netns exec {} ping -q -c 1 -W 1 {}".format(
+                                lport["name"], lport["gw"]))
+                ovs_ssh.flush()
```
hzhou8: batch mode is "False", so this line should be unnecessary?

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -487,20 +489,59 @@ def _bind_ports(self, lports, sandboxes, port_bind_args):
...
+                success = True
+                break
+            except:
+                time.sleep(0.1)
```
hzhou8: why sleep here? I makes the timeout 10% more than specified value.

hzhou8 (commented):
Sorry for the slow response. I finished the first 8 commit reviewing so far, so send the comments for now, and I will continue the rest 2 commits reviewing tomorrow.

rally_ovs/plugins/ovs/scenarios/ovn.py
dceara: @hzhou8 Ack, will be fixed in the next version.

rally_ovs/plugins/ovs/scenarios/ovn.py
dceara: @hzhou8 Right, it's not needed, i'll remove it.

rally_ovs/plugins/ovs/scenarios/ovn_sandbox.py
dceara: @hzhou8 I guess you meant `start_cidr.ip + iteration * amount + 1` right? If so, yes, I'll address it in the next version.

rally_ovs/plugins/ovs/ovnclients.py
dceara: @hzhou8 if we use the first IP to create the LRP then scenarios where `lswitch_create_args.start_cidr == "42.0.0.0/16"` would generate LRPs with IP 42.0.0.0. Some network stacks might not like that, e.g. linux `ping` considers that a broadcast IP. This will break ping tests from simulated VMs to the default gateway.

If we use the last non broadcast IP in the range as LRP address then we're fine. We'd be creating LRP with IP 42.0.254.253 and simulated VMs with IPs starting from 42.0.0.1.

On the other hand, we could say that it's invalid to configure `start_cidr == "42.0.0.0/16"` in which case we could use the old allocation scheme. But afaik we don't do such checks anywhere in the code.

I'm ok to switch back to the previous behavior if you prefer that one more.

dceara: > 
> On the other hand, we could say that it's invalid to configure `start_cidr == "42.0.0.0/16"` in which case we could use the old allocation scheme. But afaik we don't do such checks anywhere in the code.
> 
> I'm ok to switch back to the previous behavior if you prefer that one more.

@hzhou8 it's also worth to note that now with the code on master branch we would overwrite any `lswitch_create_args.start_cidr == "42.0.0.1/16"` with `lswitch_create_args.start_cidr == "42.0.0.0/16"` in `create_routed_network()`:
https://github.com/ovn-org/ovn-scale-test/blob/393bd0ff1f40e0bb911a711110a5671f100e78aa/rally_ovs/plugins/ovs/scenarios/ovn_nb.py#L44

dceara:
> Sorry for the slow response. I finished the first 8 commit reviewing so far, so send the comments for now, and I will continue the rest 2 commits reviewing tomorrow.

@hzhou8, thanks for the review! I pushed a new version with your suggested changes, except for `_get_gw_ip()` for now. There was no other change to the rest of the commits.

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -98,8 +98,15 @@ def _get_or_create_lswitch(self, lswitch_create_args=None):
...
-    def _create_lports(self, lswitch, lport_create_args = [], lport_amount=1,
-                       lport_ip_shift = 1):
+    def _create_lports(self, lswitches, lport_create_args = [], lport_amount=1,
+                       lport_ip_shift=1):
+        lports = []
+        for lswitch in lswitches:
+            lports += self._create_lport(lswitch, lport_create_args)
+        return lports
+
+    def _create_lport(self, lswitch, lport_create_args = [], lport_amount=1,
```
hzhou8: This function still creates multiple lport*s*. Would the name be misleading?

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -346,45 +357,59 @@ def _bind_ports_and_wait(self, lports, sandboxes, port_bind_args):
...
+            for lport in lports:
+                port_name = lport["name"]
+                LOG.info("bind %s to %s on %s" % (port_name, sb_name, farm))
+
+                ovs_vsctl.add_port('br-int', port_name, internal=internal)
+                ovs_vsctl.db_set('Interface', port_name,
+                                ('external_ids', {"iface-id":port_name,
+                                                  "iface-status":"active"}),
+                                ('admin_state', 'up'))
+                ovs_vsctl.flush()
```
hzhou8: Before the change, each sandbox will flush only once when len(lports) > len(sandboxes). I.e. if there are 10 ports being bound to a sandbox, they are done in one ovsdb transaction. Now with this change, it flushes for every single port binding. Is it on purpose, or shall this line be moved out of this inner for loop?

hzhou8 (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn.py
dceara: @hzhou8 Ack, I changed it to `_create_switch_lports`.

rally_ovs/plugins/ovs/scenarios/ovn.py
dceara: @hzhou8 Ack, I moved the `flush()` in the outer loop.

PR #190, opened 2020-02-21
fix bug:when use sandbox to create lports,Error NoneType object is not iterable

shenjiangc:
fix bug:when use sandbox to create lports,Error NoneType object is not iterable


huikang:
Merged as cleaned PR #191 

Thanks!

PR #195, opened 2020-05-06
Add DB clustering support for ovn-fake-multinode deployments.

dceara:


dceara:
CC: @numansiddique

igsilya (changes requested):
There are bunch of places where `self.controller_client("ovn-nbctl")` called by rally_ovs plugin, but only few of them actually calls `ovn_nbctl.set_daemon_socket(self.context.get("daemon_socket", None))` to configure socket files.  This will cause test failures if NB DB leader changed for whatever reason.

dceara:
> There are bunch of places where `self.controller_client("ovn-nbctl")` called by rally_ovs plugin, but only few of them actually calls `ovn_nbctl.set_daemon_socket(self.context.get("daemon_socket", None))` to configure socket files. This will cause test failures if NB DB leader changed for whatever reason.

@igsilya thanks for spotting this! I added a commit to address all cases and call `_get_ovn_controller()` instead of explicitly creating ovn-nbctl connections.

PR #199, opened 2020-06-01
Add the capability to attach an external command to a running test

LorenzoBianconi:
Introduce the capability to attach an external command (e.g perf or
strace) to a running ovn-scale-test for debugging purpose.
The external program can be attached before starting the workload
defining a given task in the json configuration file or providing
the start and stop iteration in the workload configuration

### DIFF
--- a/rally_ovs/plugins/ovs/scenarios/ovn.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn.py
@@ -641,3 +641,46 @@ class OvnScenario(ovnclients.OvnClientMixin, scenario.OvsScenario):
 
         ovn_nbctl = self._get_ovn_controller(self.install_method)
         return ovn_nbctl.get("Address_Set", set_name, 'addresses')
+
+    def runCmd(self, ssh, cmd="", pid_opt="", pid="", background_opt=False):
+        ssh.enable_batch_mode(False)
+
+        if pid_opt and pid:
+            cmd = cmd + " " + pid_opt + " " + pid
+
+        if background_opt:
+            cmd = cmd + " &"
+
+        stdout = StringIO()
+        ssh.run(cmd, stdout=stdout)
+        return stdout.getvalue().rstrip()
+
+    def runFarmCmd(self, sandbox, cmd="", pid_opt="", pid_proc_name="",
+                   background_opt=False):
+        LOG.info("running %s on %s" % (cmd, sandbox["name"]))
+
+        ssh = self.farm_clients(sandbox["farm"], "ovs-ssh")
+        ssh.set_sandbox(sandbox["name"], self.install_method,
+                        sandbox["host_container"])
+
+        if pid_proc_name:
+            pid = self.runCmd(ssh, "pidof -s " + pid_proc_name)
+        else:
+            pid = ""
+
+        self.runCmd(ssh, cmd, pid_opt, pid, background_opt)
+
+    def runControllerCmd(self, cmd="", pid_opt="", pid_proc_name="",
+                         background_opt=False):
+        LOG.info("running %s on controller-sandbox" % cmd)
+
+        ssh = self.controller_client("ovs-ssh")
+        ssh.set_sandbox("controller-sandbox", self.install_method,
+                        self.context["controller"]["host_container"])
+
+        if pid_proc_name:
+            pid = self.runCmd(ssh, "pidof -s " + pid_proc_name)
+        else:
+            pid = ""
+
+        self.runCmd(ssh, cmd, pid_opt, pid, background_opt)
--- a/rally_ovs/plugins/ovs/scenarios/ovn_nb.py
+++ b/rally_ovs/plugins/ovs/scenarios/ovn_nb.py
@@ -226,7 +226,8 @@ class OvnNorthbound(ovn.OvnScenario):
                             port_bind_args = None,
                             create_acls = True,
                             name_space_size = 1,
-                            network_policy_size = 1):
+                            network_policy_size = 1,
+                            ext_cmd_args = {}):
         lswitches = self.context["ovn-nb"]
         ip_offset = lport_create_args.get("ip_offset", 1) if lport_create_args else 1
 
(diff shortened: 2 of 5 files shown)


rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -639,3 +639,34 @@ def _get_address_set(self, set_name):
 
         ovn_nbctl = self._get_ovn_controller(self.install_method)
         return ovn_nbctl.get("Address_Set", set_name, 'addresses')
+
+    def runCmd(self, ssh, cmd="perf record", process="ovn-controller"):
```
dceara: This seems very `perf` specific. Also, what if there are multiple processes with name `process` running at the same time. `pidof` will only select one.

Also, why do we need a special treatment for "stop" commands? Can't we have something like below?
``runCmd(ssh, 'perf record -p `pidof ovn-controller`')`` and later ``runCmd(ssh, 'kill `pidof ovn-controller`)''

rally_ovs/plugins/ovs/scenarios/ovn_nb.py
```diff
@@ -226,19 +226,38 @@ def create_routed_lport(self, lport_create_args = None,
                             port_bind_args = None,
                             create_acls = True,
                             name_space_size = 1,
-                            network_policy_size = 1):
+                            network_policy_size = 1,
+                            ext_cmd_args = {}):
```
dceara, line 230: Why do we enclose the two args in a dict here? We already pass a lot of args one by one, I think it's a bit more clear if we don't have a dict() that contains the last two arguments for a scenario.

rally_ovs/plugins/ovs/scenarios/ovn_nb.py
```diff
@@ -226,19 +226,38 @@ def create_routed_lport(self, lport_create_args = None,
...
+               self.handle_cmd(stop_cmd)
+
         self.configure_routed_lport(lswitch, lport_create_args,
                                     port_bind_args, ip_start_index,
                                     name_space_size, network_policy_size,
                                     create_acls)
 
+    @scenario.configure()
+    def handle_cmd(self, cmd_args = {}):
+        sandbox_size = cmd_args.get("size", 0)
```
dceara: Maybe we can find a better name than "size", what do you think? By looking at the json file it's not so clear that it refers to the number of sandboxes to run the command on.

LorenzoBianconi: @dceara ack, I have not too imagination actually. What about `num_sanboxes` ?

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: > This seems very `perf` specific. Also, what if there are multiple processes with name `process` running at the same time. `pidof` will only select one.

@dceara why 'perf' specific? at the end 'perf' it is just used as default cmd value, right?
Yes, in theory we can have multiple process in the same sandbox but does it worth to complicate the code now for an out of scope case?

> 
> Also, why do we need a special treatment for "stop" commands? Can't we have something like below?
> `` runCmd(ssh, 'perf record -p `pidof ovn-controller`') `` and later ``runCmd(ssh, 'kill `pidof ovn-controller`)''

@dceara yes, I think it works, I will update it

rally_ovs/plugins/ovs/scenarios/ovn_nb.py
LorenzoBianconi: @dceara I think it is more clear if we pass all the arguments related to ext-cmd in a single block, no? maybe we can think to re-factor the rest of the code

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2016 Ebay Inc.
+#pick Copyright 2016 Ebay Inc.
```
dceara: Typo during rebase?

rally_ovs/plugins/ovs/scenarios/ovn.py
```diff
@@ -639,3 +639,42 @@ def _get_address_set(self, set_name):
...
+
+    def runFarmCmd(self, sandbox, cmd="", pid_opt="", background_opt=False):
+        LOG.info("running %s on %s" % (cmd, sandbox["name"]))
+
+        ssh = self.farm_clients(sandbox["farm"], "ovs-ssh")
+        ssh.set_sandbox(sandbox["name"], self.install_method,
+                        sandbox["host_container"])
+
+        if pid_opt:
+            pid = self.runCmd(ssh, "pidof -s ovn-controller")
```
dceara: What if we want to run a command like:
 ```
perf record -p `pidof -s ovs-vswitchd`
```

I think we need another `pid_proc_name` argument, or something like that.

File: rally_ovs/plugins/ovs/scenarios/ovn.py [hunk 6]
```diff
@@ -639,3 +639,42 @@ def _get_address_set(self, set_name):
...
+
+    def runControllerCmd(self, cmd="", pid_opt="", background_opt=False):
+        LOG.info("running %s on controller-sandbox" % cmd)
+
+        ssh = self.controller_client("ovs-ssh")
+        ssh.set_sandbox("controller-sandbox", self.install_method,
+                        self.context["controller"]["host_container"])
+
+        if pid_opt:
+            pid = self.runCmd(ssh, "pidof -s ovn-northd")
```
dceara: Same here, what if we want to run a command like:
 ```
perf -p `pidof -s ovn-nbctl`
```

I think we need another `pid_proc_name` argument, or something like that.

File: rally_ovs/plugins/ovs/scenarios/ovn.py [hunk 7]
```diff
@@ -639,3 +639,42 @@ def _get_address_set(self, set_name):
 
         ovn_nbctl = self._get_ovn_controller(self.install_method)
         return ovn_nbctl.get("Address_Set", set_name, 'addresses')
+
+    def runCmd(self, ssh, cmd=""):
```
dceara: I think we should add the `background_opt` argument here instead of doing the same work twice in `runFarmCmd` and `runControllerCmd`. Same for `pid_opt` (and potentially `pid_proc_name` if we add it - see below).

rally_ovs/plugins/ovs/scenarios/ovn_nb.py
dceara: Ok.

dceara (changes requested):

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: @dceara I think so :)

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: @dceara ack, fine I will update it

rally_ovs/plugins/ovs/scenarios/ovn.py
LorenzoBianconi: @dceara ack, I will update it

dceara (approved):
Looks good to me, thanks!

PR #201, opened 2020-06-27
Fix missing args when calling _create_switch_lports().

hzhou8:
This causes the scenarios such as create_routers_bind_ports fail.

Signed-off-by: Han Zhou <hzhou8@ebay.com>


dceara (approved):
Looks good to me, thanks!

PR #202, opened 2020-06-30
Fix ovn-sbctl calls that might hang in a clustered DB scenario.

dceara:
ovn_fake_multinode scenarios check if a chassis has registered with the
SB DB by running ovn-sbctl commands on the first central instance.
However, if a leadership change happens and the first central instance
is not SB DB leader anymore, the ovn-sbctl call will hang.

To avoid such issues we add the "--no-leader-only" option to ovn-sbctl
calls in this case.

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -421,7 +421,7 @@ class OvnSbctl(OvsClient):
             self.run("sync", opts)
             self.batch_mode = batch_mode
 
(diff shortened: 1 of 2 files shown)


numansiddique:
Acked-by: Numan Siddique <numans@ovn.org>

dceara:
Abandoning this change, "--no-leader-only" is already passed to the sbctl command:
https://github.com/ovn-org/ovn-scale-test/commit/c3d3b4616ca30520eafe754e68a95b113a7933d4#diff-1834c220ccdc0f1b1966eb6597b1edb5R329

PR #204, opened 2020-07-15
install platform-python-devel on rhel as python-devel doesn't exit

linup2011:
Signed-off-by: jishi <jishi@redhat.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -280,7 +280,16 @@ install_required_sw () {
         fi
     elif have_command yum; then
         # RHEL/CentOS
-        missing=$(which_missing_packages gcc libffi-devel python-devel openssl-devel gmp-devel libxml2-devel libxslt-devel postgresql-devel redhat-rpm-config git wget)
+        missing=$(which_missing_packages gcc libffi-devel openssl-devel gmp-devel libxml2-devel libxslt-devel postgresql-devel redhat-rpm-config git wget)
+        if yum info python-devel
+        then
+            missing="$missing $(which_missing_packages python-devel)"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -280,7 +280,15 @@ install_required_sw () {
         fi
     elif have_command yum; then
         # RHEL/CentOS
-        missing=$(which_missing_packages gcc libffi-devel python-devel openssl-devel gmp-devel libxml2-devel libxslt-devel postgresql-devel redhat-rpm-config git wget)
+        if yum info python-devel
```
dceara, line 284: This works but there's quite a bit of duplication.

Could you please move the common "missing" packages here, before the `if`? I think it's better to add to `$missing` inside the if based on the availability of the `python-devel`/`platform-python-devel` packages. Maybe something like:

```
missing=$(which_missing_packages gcc libffi-devel openssl-devel gmp-devel libxml2-devel libxslt-devel postgresql-devel redhat-rpm-config git wget)

if yum info python-devel
then
    missing="$missing $(which_missing_packages python-devel)"
elif yum info platform-python-devel
    missing="$missing $(which_missing_packages platform-python-devel)"
else
    missing="$missing $(which_missing_packages python-devel)"
fi
```

dceara (changes requested):
Just a style related comment but I think it would make the code a bit more readable.

Thanks,
Dumitru

install.sh
linup2011: updated

dceara (approved):

PR #207, opened 2020-08-24
Misc improvements

hzhou8:
1. Change script file to executable after deploying on farm node.
2. Support separate OVS and OVN for post 2.12 sandbox installation.
3. sandbox: Disable ovn-northd NB/SB probe.
4. sandbox: Disable ovn-controller ofctrl openflow probe.
5. sandbox: Set OVS n-xxx-threads to 1


l8huang (approved):
LGTM

PR #209, opened 2020-08-25
sandbox: Fix ovs-vsctl batch mode.

hzhou8:
ovs-vsctl batch mode was not real batch mode. It still executes multiple
commands (e.g. for port binding) one by one. This patch fixes it with real
batch mode that executs all commands in a single transaction.

Note: this works only for sandbox mode. For using docker, it still executes
commands separately.

Signed-off-by: Han Zhou <hzhou8@ebay.com>

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -525,8 +525,13 @@ class OvsVsctl(OvsClient):
             self.cmds = self.cmds or []
 
             # TODO: tested with non batch_mode only for docker
-            if self.install_method == "docker":
+            if self.install_method == "docker" or self.host_container:
                 self.batch_mode = False
+
+            if self.batch_mode:
+                cmd = itertools.chain([" -- "], opts, [cmd], args, extras)
+                self.cmds.append(" ".join(cmd))
+                return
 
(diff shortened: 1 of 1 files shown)


rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -556,11 +558,13 @@ def flush(self):
             if self.cmds == None:
                 return
 
+            run_cmds = []
             if self.sandbox:
                 if self.install_method == "sandbox":
-                    self.cmds.insert(0, ". %s/sandbox.rc" % self.sandbox)
+                    run_cmds.append(". %s/sandbox.rc" % self.sandbox)
+                    run_cmds.append("ovs-vsctl" + " ".join(self.cmds))
```
l8huang, line 565: looks like a whitespace is needed after "ovs-vsctl" 

```
run_cmds.append("ovs-vsctl " + " ".join(self.cmds))
```

hzhou8: It works in my environment. I think it is because a whitespace is already added for batch mode at L532: cmd = itertools.chain([" -- "], opts, [cmd], args, extras)

l8huang (approved):
LGTM

PR #211, opened 2020-08-26
ovn.py: Add ovn_network.bind_port internal_vm argument.

dceara:
This controls whether the port should be simulated by a network
namespace.

Default value is "true" to maintain compatibility with older versions.


putnopvut (approved):
Acked-by: Mark Michelson <mmichels@redhat.com>

PR #213, opened 2020-11-17
ovs_client: introduce option to log commands

LorenzoBianconi:
Introduce log_cmd option to log ovn-nbctl, ovn-sbctl, ovs-ssh, ovs-vsctl
and ovs-ofctl commands

### DIFF
--- a/rally_ovs/plugins/ovs/ovsclients_impl.py
+++ b/rally_ovs/plugins/ovs/ovsclients_impl.py
@@ -36,7 +36,7 @@ class SshClient(OvsClient):
 class OvnNbctl(OvsClient):
 
 
-    class _OvnNbctl(DdCtlMixin):
+    class _OvnNbctl(DdCtlMixin, OvsClientLogger):
         def __init__(self, credential):
             self.ssh = get_ssh_from_credential(credential)
             self.is_root = is_root_credential(credential)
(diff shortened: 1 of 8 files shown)


rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -125,7 +131,8 @@ def flush(self):
 
                     run_cmds.append("{} {} {}".format(cmd_prefix, ovn_cmd,
                                                       " ".join(self.cmds)))
-
+            if self.log_cmd:
+                LOG.info("".join(run_cmds))
```
dceara: IIUC this should be `"\n".join(run_cmds)`.  This applies to multiple places in this PR.

Also, isn't it simpler if we create a base logging class in `ovsclients.py`?  Something like below:

```
from rally.common import logging

[...]

LOG = logging.getLogger(__name__)

def configure(name):
    return plugin.configure(name, namespace=_NAMESPACE)

class OvsClientLogger(object):
    def set_log_cmd(self, log_cmd=False):
        self.log_cmd = log_cmd

    def log_cmds(self, cmds):
        if self.log_cmd:
            LOG.info(cmds);
```

Then all client classes can inherit from this, i.e.:
```
class _OvnNbctl(DdCtlMixin, OvsClientLogger)
class _OvnSbctl(DdCtlMixin, OvsClientLogger)
class _OvsSsh(OvsClientLogger)
class _OvsVsctl(OvsClientLogger)
class _OvsOfctl(OvsClientLogger)
```

dceara (changes requested):
Overall this looks OK but I'd suggest refactoring the change a bit and the logging of batched commands should be fixed.

Thanks!

dceara (approved):
LGTM, thanks!

PR #214, opened 2021-02-12
Add new scenario ovn_run_command().

dceara:
This allows running custom commands on the test cluster, e.g., enable DP
groups or configure RAFT election timeouts.


numansiddique (approved):
LGTM

PR #215, opened 2021-02-19
Enable batch mode for create_routed_lport

LorenzoBianconi:
Introduce the capability to run create_routed_lport in batch mode in
order to reduce the time needed to carry out the workload.


dceara (approved):
LGTM, thanks!

PR #218, opened 2021-06-03
[WIP] Add load balancer support.

dceara:


dceara:
@putnopvut  It would be great if you could have a look at this, thanks!

rally_ovs/plugins/ovs/scenarios/ovn_fake_multinode.py
```diff
@@ -418,6 +418,11 @@ def setup_switch_per_node(self, fake_multinode_args = {},
                               lport_create_args = {},
                               port_bind_args = {},
                               create_mgmt_port = True):
+        # TODO: figure out how to not reload the context?
```
dceara, line 421: @putnopvut This is the hack I was talking about.

rally_ovs/plugins/ovs/ovsclients.py
```diff
@@ -172,6 +172,29 @@ def parse_lswitch_list(lswitch_data):
...
+        else:
+            tokens = line.split('=')
+            if len(tokens) == 0:
+                lb_name = None
+                continue
+            lb_vip = tokens[0]
+            lb_backends = tokens[1] if len(tokens) == 2 else ''
+            lbs[lb_name] = (lb_vip, lb_backends)
+            lb_name = None
+
```
putnopvut, line 195: If the `=` token is not found, then `split()` will return a list containing the whole string as the first item. Therefore, `len(tokens)` wont' ever be 0.

You can just replace this whole block with
```
lbs[lb_name] = tuple(line.split('=', 2))
lb_name = None
```
This will ensure that `lbs[lb_name]` is set to a tuple of 2 items. If the `=` is not present, then the second item of the tuple will be an empty string.

rally_ovs/plugins/ovs/ovsclients_impl.py
```diff
@@ -282,6 +282,26 @@ def port_group_del(self, port_group):
             params = [port_group]
             self.run("pg-del", [], params)
 
+        def lb_add(self, lb_name, lb_vip, lb_proto):
+            params = [lb_name, lb_vip, '\'\'', lb_proto]
+            self.run("lb-add", [], params)
+        
+        def lb_set_vip_backends(self, lb_name, lb_vip, lb_backends):
+            vips_set='vips=\'\"{}\"=\"{}\"\''.format(lb_vip, lb_backends)
```
putnopvut, line 290: Nit, you can get rid of the `\` before the double quotes. It might make the line a bit easier to read.

rally_ovs/plugins/ovs/scenarios/ovn_fake_multinode.py
putnopvut: Can you go into a bit more detail about what's going on here? I've been away from this for long enough that I don't even know what it means to reload the context.

dceara: `self.context` gets populated once before any iteration of the scenario is run, in `class OvnNorthboundContext.setup()`.  Then, if I'm not wrong, each scenario gets a copy of the context.  However, because we're updating load balancers in each iteration, ideally we would have to update the original context, such that upcoming iterations have the updated set of load balancers.

The rally runner doesn't allow that because iterations might be run in parallel, I presume.

To work around it I explicitly updated self.context["ovn-nb-lbs"] with the results of `ovn_nbctl.lb_list()`.

rally_ovs/plugins/ovs/ovsclients_impl.py
dceara: Ack, thanks.

rally_ovs/plugins/ovs/ovsclients.py
dceara: Way better, will do, thanks!

